PR #691, opened 2015-01-24
Update sws_osx_install.command to be more robust.

cowboy:
* Fails with a message if `~/Library/Application Support/REAPER` cannot be found; Eg. if the user has a portable installation.
* Allows the user to specify the destination directory from the command-line.
* Removed unnecessary duplication.


Jeff0S:
Sorry for the late feedback: I was planning to re-do the OSX installer (which is well convoluted ATM), and get rid of this script on the way but this will have to wait a bit more... So your contribution is welcome, thanks!

Would you mind re-doing this pull request on the "next" branch though?
(could merge it by hand, but it would be better to keep track of the original author...)

cowboy:
Oh crap. Let me close this and do it over. Sigh.

PR #708, opened 2015-04-18
Update sws_osx_install.command to be more robust.

cowboy:
* Fails with a message if `~/Library/Application Support/REAPER` cannot be found; Eg. if the user has a portable installation.
* Allows the user to specify the destination directory from the command-line.
* Removed unnecessary duplication.

Redo of #691 against `next` branch per @Jeff0S.


Jeff0S:
Thank you cowboy!

PR #812, opened 2016-05-20
Autocolor: "0x" prefix added automatically.

rokit:
Users can now enter 6-digit hexadecimal values and the "0x" prefix will be added automatically.  This is useful for people like me who like to copy 6-digit hex values from the web, Photoshop, etc, and paste them.

### DIFF
--- a/Color/Autocolor.cpp
+++ b/Color/Autocolor.cpp
@@ -181,7 +181,16 @@ void SWS_AutoColorView::SetItemText(SWS_ListItem* item, int iCol, const char* st
 				break;
 			case COL_COLOR:
 			{
-				int iNewCol = strtol(str, NULL, 0);
+				char newstr[] = "0x"; // common prefix, so the user doesn't have to type it themselves -Kochab
+				int iNewCol = -1;
+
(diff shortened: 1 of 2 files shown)


swstim:
Hey rokit.  I can't merge this as is as there's a potentially nasty bug - you never actually allocate any memory for newstr (`char newstr[]` just gives you a pointer to character string) and strcat will just append onto unallocated memory.

Besides there's a much better way:
```c
    int iNewCol;
    if (strlen(str) == 6)
        iNewCol = strtol(str, NULL, 16); // force base 16
    else
        iNewCol = strtol(str, NULL, 0); // auto detect base from prefix
````

Not to mention I'm in favor of just making it force b16 all the time, so just making a 2 character change:
```
- int iNewCol = strtol(str, NULL, 0);
+ int iNewCol = strtol(str, NULL, 16);
```

rokit:
Thanks for the explanation.  I wasn't aware of the memory issue.  I always heard C++ could trip people up in that area and I'm not too versed in it.  I obviously have some reading to do as far as understanding base(n) and bitwise operators as well.  I'm glad you like the idea nonetheless.

PR #844, opened 2017-03-17
Remove old, useless and likely very buggy Xenakios envelope generatio…

Xenakios:
…n action


swstim:
Cherry picked into next branch (so as to keep taglib for vs12 build)

Issue #781, opened 2015-11-29
(FR) API: Get Item Max Peak and Average RMS Level


X-Raym:
Hi,

SWS is capable of analyzing items and determine what is it's Max Peak level, and what is it's Average RMS level. This function is use in the SWS label processor and in the Organize Items by Peak/RMS actions.

It could be nice if we could get this value from ReaScript, with two functions like this:
max_peak = GetMediaItemMaxPeak( item )
average_rms = GetMediaItemAverageRMS( item )

Currently, we can only access this value by using the SWS label processor, and renaming all the items with the level we want. No very flexible.

Thanks for listening !

nofishonfriday:
I'm assigning to this request.
No promises, but I have it basically working already. Will sprobably send a pull request for review to Tim soon.

X-Raym:
Added in v2.9.3 thanks to @nofishonfriday  !

PR #856, opened 2017-03-30
implement #781

nofishonfriday:
This exports two functions to ReaScript:

reaper.NF_GetMediaItemMaxPeak()
reaper.NF_GetMediaItemAverageRMS()

Lua Testscript::

```
reaper.ShowConsoleMsg("")

function msg(m)
  return reaper.ShowConsoleMsg(tostring(m) .. "\n")
end


item = reaper.GetSelectedMediaItem(0, 0)

maxPeak = reaper.NF_GetMediaItemMaxPeak(item) 
avrgRMS = reaper.NF_GetMediaItemAverageRMS(item)


msg(maxPeak)

msg("\n")

msg(avrgRMS)
```

I've also updated the ReaScript doc.
Tim, please see my comments in Analysis.cpp, there's certainly room for improvement.

Thanks.

### DIFF
--- a/Breeder/BR_ReaScript.cpp
+++ b/Breeder/BR_ReaScript.cpp
@@ -35,6 +35,9 @@
 #include "../SnM/SnM_Chunk.h"
 #include "../SnM/SnM_Util.h"
 
+// #781
+#include "../Misc/Analysis.h"
+
 /******************************************************************************
 * Globals                                                                     *
 ******************************************************************************/
@@ -918,3 +921,17 @@ bool BR_Win32_WritePrivateProfileString (const char* sectionName, const char* ke
 {
 	return !!WritePrivateProfileString(sectionName, keyName, value, filePath);
 }
+
+// #781
+double NF_GetMediaItemMaxPeak(MediaItem* item)
+{
+	double maxPeak = GetMediaItemMaxPeak(item);	
+	return maxPeak;
+}
+
+double NF_GetMediaItemAverageRMS(MediaItem* item)
+{
+	double avrgRMS = GetMediaItemAverageRMS(item);
+	return avrgRMS;
+}
+
--- a/Breeder/BR_ReaScript.h
+++ b/Breeder/BR_ReaScript.h
@@ -96,6 +96,10 @@ int             BR_Win32_GetPrivateProfileString (const char* sectionName, const
 int             BR_Win32_ShellExecute (const char* operation, const char* file, const char* parameters, const char* directoy, int showFlags);
 bool            BR_Win32_WritePrivateProfileString (const char* sectionName, const char* keyName, const char* value, const char* filePath);
 
+// #781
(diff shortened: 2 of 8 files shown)


nofishonfriday:
I added https://github.com/reaper-oss/sws/pull/856/commits/098cf7382da67a4f76562f75a5967dc5c46f648e as a separate commit (changed action naming from nofish to NF for consistency with the other SWS actions).

swstim:
Good enough for me!

X-Raym:
@nofishonfriday Just get your PM, and tested your functions,

seems nice to me !

Would it be possible to prevent the popup analysis window to appear ?
Screenshot

Also, rather than ITEM, I think this should be a TAKE api, so that we can check only the active take, which would be more flexible. Sorry, I didn't think about this issue when I wrote this Fr :/

Thanks for your work on that !

X-Raym:
@nofishonfriday just edited my post above,
I realize that the function already works with Active Take and Active Channel, which is the desired behavior.
Also, it work post gain, which is cool !

So, the only issue is the analysis window. :)

nofishonfriday:
Hi X-Raym,

I also thought about the appearing analysis window when doing it, and I assumed Tim (or whoever did the original item analysis functions, which these actions are just re-using) made that progress windows for a reason.

So I'd rather refrain in this case from changing the original analysis functions (someone else's code) and leave this question to Tim (you might ask him directly if you want :) who probably knows more about it than me, sorry.

X-Raym:
@nofishonfriday Ah yes two things,
would it be possible to make it return "invalid source" or any error message instead of "-150" ? Would be a bit more flexible :)

For the API doc,
instead of
`Returns the greatest max. peak value of all channels of an audio item. Returns -150.0 if MIDI item or empty item.`
A more complete doc can be
`Returns the greatest max. peak value of active channels of an audio item active take, post item gain, post fader, post take volume envelope, post-fade, pre item FX. Returns -150.0 if MIDI item or empty item.`

X-Raym:
@nofishonfriday ok the analysis window ! The problem with it is that it is so fats that we don't really understand what is happening ^^ I guess it is requested for very long items so that we understand why REAPER is hanging. So, don't worry for that pop up :)

nofishonfriday:
> would it be possible to make it return "invalid source" or any error message instead of "-150" ? Would be a bit more flexible `:)`

The way  it's now is  also (partly) based in re-using the original SWS item analysis function. Try with a MIDI item, it'll will also return -150.0 P:

Ok, I know it's not the best thing to handle it, will think about it. :)

Thanks for the API doc additions, this will be in.

X-Raym:
@nofishonfriday maybe a fix can be proposed right inside the analysis function. I don't think having -150 as peak level for midi (with label processing for eg, which also uses this analyses function) is that useful.
But as the MIDI/Empty item check can be perform before the analyses function, don't worry too much about that ! it is just perfectionism. ^^

Issue #945, opened 2018-01-24
Include Latch-Preiview in the action "SWS/S&M: Toolbar - Toggle track envelopes in touch/latch/write"


robingshore:
It would be very helpful if tracks that are in Latch-Preview automation mode were toggled by the action "SWS/S&M: Toolbar - Toggle track envelopes in touch/latch/write". Right now this action only affects tracks that are in touch, latch, and write mode.

nofishonfriday:
merged.

PR #948, opened 2018-02-04
"Toggle track envelopes in touch/latch/write" - add latch preview (#945)

nofishonfriday:


nofishonfriday:
Currently it states on SWS site "Requires REAPER v5.21 or above" but Latch preview mode was added in v5.22, so I've put in a check for Reaper version in. If required version will be bumped to 5.22 I could remove the check.

PR #979, opened 2018-05-09
(dirty) re-fix BR_SetItemEdges() (#950)

### LINKED ISSUE #950: Bug: BR_SetMediaItemEdges should keep envelope points in place
Hi,

The ReaScript function BR_SetMediaItemEdges keep envelope points position relative to the item left edges, instead as leaving them at their absolute position in the project (the whole point of this function is to modify items and by preserving absolute position of the content, so envelope points should move).

Steps:

1. Have an item
2. Create an envelope
3. Add points
4. use the SetMediaEdges function
5. see that envelope has been offset.

demo

Thanks for taking a look :)

nofishonfriday:
https://github.com/reaper-oss/sws/issues/950#issuecomment-384251678 ff

fixes #950

### DIFF
--- a/Breeder/BR_Util.cpp
+++ b/Breeder/BR_Util.cpp
@@ -1147,13 +1147,38 @@ bool TrimItem_UseNativeTrimActions(MediaItem* item, double start, double end, bo
 	double itemLen = GetMediaItemInfo_Value(item, "D_LENGTH");
 
 	if (force || start != itemPos || newLen != itemLen) {
+
+		// NF: ugly code ahead for fixing
+		// https://github.com/reaper-oss/sws/issues/950#issuecomment-384251678
+		// but it avoids having to deal with stretch markers (see other comments in #950)
+
+		// save item selection
+		WDL_TypedBuf<MediaItem*> selItems;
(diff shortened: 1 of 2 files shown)


X-Raym:
@nofishonfriday Hi !

Now that we have a test build, I experimented more with this long awaited function.

and I found a bug in this PR !

In fact, item grouping has to be deactivated and restored from the function itself, else it edit all the items in the group.

Here is a script which work on only one selected item.

demo

```lua
item = reaper.GetSelectedMediaItem(0,0)

if not item then item = reaper.GetMediaItem(0,0) end

reaper.BR_SetItemEdges(item, 16,17)

reaper.UpdateArrange()
```

can you fix this for #1092 ? Thx ! 😄

X-Raym:
@nofishonfriday tested and it works fine!

@cfillion @nofishonfriday so... are we good for proposing this PR ? 😈

nofishonfriday:
@X-Raym 
Depends on if cfillion plans to add something for #1086 I think.
If not, I'd suggest mentioning it as known issue in the whatnew.

PR #996, opened 2018-05-30
func: get vertical position of view in arrange area

doppgit:


doppgit:
removed 2 actions from Misc/TrackParams.cpp, which i did previously, because the same result can be achieved with functions provided by juliansader in pull request #1028. I don't see a point to add the same functionality.

PR #1028 also contains function, which can replace my DO_GetArrangeVertPos, but i still would like to add DO_GetArrangeVertPos to SWS as an alternative, because it's smaller, returns only 2 parameters, which can be more efficient for some specific defered scripts, like mine for example https://github.com/doppgit/spells/blob/master/dopp_fxList_public.lua. 
Thanks! Hope to see update soon.

doppgit:
all i did here is not needed anymore, my script used this functions, but looks like script is useless to other people, and this functions can be replaced by functions from other pull request1028, so to avoid redundancy i remove pull request.

PR #1026, opened 2018-07-02
Window and Mouse functions

juliansader:
Added a bunch of window- and mouse-related functions to expose standard Win32/swell functions to API.


doppgit:
useful stuff, i planned to make few similar functions, that you already made=), so i'm stopping, the only thing, that i don't understand, why you add all that functions in reascript_vararg.h, isn't it will be automatically created, when extension will be built?

juliansader:
The new stuff in reascript_vararg.h were indeed automatically created when I built the extension.

I'm not sure which files I should upload for the pull request, so I tried to upload all files that were changed.

The important files are
* js_swell_ReaScript.cpp
* js_swell_ReaScript.h
* ReaScript.cpp (with the new API function descriptions)
* sws_extensions.cpp (with two additions to load Get/SetExtState)

juliansader:
I am closing this pull request and will open a new one for the "next" branch.

PR #1028, opened 2018-07-04
Next: Added some Window and Mouse API functions

juliansader:
(I closed the "master" version of this pull request, and opened this request for "next".)

Added a bunch of window- and mouse-related functions to expose standard Win32/swell functions to API.

Whatsnew.txt has also been updated with the list of new functions.


juliansader:
I hope version 2.9.8 of SWS can be released soon!

juliansader:
Added a few functions to load and set mouse cursor.

juliansader:
Added Window_PostMessage

juliansader:
Added functions to intercept mousewheel messages.

mespotine:
Fantastic works and really helpful functions.

I toyed around with them for a few days and want to give some feedback on them, though.

a) the functionnames should contain a "signature". Means, the name of function Window_HandleFromAddress() would become JS_Window_HandleFromAddress(), etc. That way, you can prevent possible naming-conflicts with reaper's future-API-functionnames.

b) Window_GetTitle() doesn't return the title of the window, but rather the title of the currently focused gui-element, which may be the window, but also buttons, checkboxes, lists, etc. 
This is actually a great thing and I would love to suggest some more additions, if that is possible:
- returning the title of the window itself(which is currently missing)
- the type of the gui-element, maybe it's position in the window(x and y)
This would allow to know exactly, what element is focused in a window, giving us even more chances to signal to the user, where the current focus lies.

So if a button(like a Ok Button in a dialogbox) is focused, we could display "Ok, Button is focused, when you hit space-key" or something like that. That would give us a heck lot of use-cases :)

c) In the docs: "For more information on message codes, refer to the Win32 C++ API documentation." it would be cool, if you could add an URL to such a website. I'm not sure whether something like that exists, but it would be cool to have that.

The mousefunctions are really great, especially Mouse_GetState. Is there a chance to add doubleclick and dragging to it as well?

cfillion:
Justin recently patched upstream SWELL's Linux FindWindowEx justinfrankel/WDL@502e352912b7b216bb3c4f776a41700bf48a62cc.

juliansader:
> b) Window_GetTitle() doesn't return the title of the window, but rather the title of the currently 
> focused gui-element, which may be the window, but also buttons, checkboxes, lists, etc.
Window_GetTitle(w) should -- if everything is working OK -- return the title of w, whether or not w is focused.  w may be a child window or even a tiny GUI element.


> returning the title of the window itself(which is currently missing)
To get the title of the window, you can call Window_GetParent until you reach the desired level in the hierarchy.  For example, to get the MIDI editor, call GetParent until reaper.MIDIEditor_GetMode(w) ~= -1.


> the type of the gui-element, maybe it's position in the window(x and y)
To get the x,y position of windows and GUI elements, you can use GetRect or GetClientRect.  
I'll have to check whether it is possible to get more information about the element type.


> This would allow to know exactly, what element is focused in a window, 
> giving us even more chances to signal to the user, where the current focus lies.
Also try Window_GetFocus().


> c) In the docs: "For more information on message codes, refer to the Win32 C++ API
> documentation." it would be cool, if you could add an URL to such a website. I'm not 
> sure whether something like that exists, but it would be cool to have that.
I will add some links.  Microsoft itself has lots of documentation, for example:
`https://docs.microsoft.com/en-us/windows/desktop/learnwin32/window-messages`.


> The mousefunctions are really great, especially Mouse_GetState. 
> Is there a chance to add doubleclick and dragging to it as well?
Mouse_GetState is limited to the states returned by GetAsyncKeyState, which does not include double click and dragging. However, double click can tracked by checking whether the state quickly changed from ON to OFF and back to ON again within a certain time limit.

The WindowMessage functions can be used for more advanced mouse functions.  For example, to check double click, you can start intercepting double-click messages with:
WindowMessage_Intercept(window, "WM_LBUTTONDBLCLK:passthrough")
and then check for messages in each defer cycles with:
WindowMessage_Peek(window, "WM_LBUTTONDBLCLK")

For dragging, check for WM_LBUTTONDOWN and then WM_MOUSEMOVE.

PR #1070, opened 2018-12-03
FNG Groove tool: Add action "SWS/FNG: Apply selected groove(...)"

nofishonfriday:
Adds action (Main section):
SWS/FNG: Apply selected groove (use curent settings from opened groove tool)

This topic came up [here](https://forum.cockos.com/showpost.php?p=2065352&postcount=2496).
It's not 100% what user requests (because Groove tool must be open for the new action to work), but the problem is the Groove tool doesn't retain selected groove on closing/reopening (see gif below) and I don't know how to change this currently, so I think this action is at least better than nothing for now.
If someone comes up with a better approach, always welcomed.

[gif](https://user-images.githubusercontent.com/7658194/49347247-e4472900-f69c-11e8-84e3-044fd94dc294.gif)

### DIFF
--- a/Fingers/GrooveDialog.h
+++ b/Fingers/GrooveDialog.h
@@ -6,6 +6,7 @@ class GrooveDialog : public SWS_DockWnd
 public:
 	GrooveDialog();
 	void Refresh() { if (IsValidWindow()) RefreshGrooveList(); }
+	void ApplySelectedGroove(); // NF: made public, so we can call from action list
 	virtual ~GrooveDialog();
 private:
 	// SWS_DockWnd overrides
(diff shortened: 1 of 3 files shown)


Fingers/GrooveDialog.h
```diff
@@ -6,6 +6,7 @@ class GrooveDialog : public SWS_DockWnd
 public:
 	GrooveDialog();
 	void Refresh() { if (IsValidWindow()) RefreshGrooveList(); }
+	void ApplySelectedGroove_public(); // NF: so we can call from action list
```
cfillion: Couldn't `ApplySelectedGroove` be made public directly?

nofishonfriday: Thanks for the review.
> Couldn't ApplySelectedGroove be made public directly?

That's what I had as original version indeed but then I thought maybe it's not that good style making the whole function public which was private originally, so I added a 'dedicated' public version instead.
But you're right of course, practically makes no different, I just wasn't sure about 'coding style'.

nofishonfriday: Implemented suggested change in https://github.com/reaper-oss/sws/commit/60cdaa9d73974164c2d92d4649368abe8fab2b69

PR #1100, opened 2019-02-04
'Disable / Enable multichannel metering' actions: improve performance

nofishonfriday:
report / discussion here:
https://forum.cockos.com/showthread.php?p=2090523#post2090523
\+ small Auto group tweaks


swstim:
I think the performance hit is not the parsing itself, it's passing a new chunk back to reaper - it doesn't know what exactly you've changed so it has to update the entire track state.  In a chunk with plugins that can get expensive fast.

swstim:
Can you update the whatsnew.txt please?

swstim:
You're welcome.  Now that there's not a huge intimidating pile of PRs to look at I should be able to address these faster.

PR #1124, opened 2019-03-10
Resources: set FX offline if 'Avoid loading undo states when possible' is enabled before calling SetTrackStateChunk()

### LINKED ISSUE #1109: Resource Window: "Apply track template to selected tracks" not working correctly
Had to keep the topic short, sorry.

The issue is: when a track already has Kontakt loaded and you click on "apply to selectes tracks", the track template (also consisting of a Kontakt instance) is not loaded, because Kontakt is already in the FX slot. However, the expected behavior would be to load the track template in any case.

nofishonfriday:
(to force updating FX settings)
fixes #1109

The code is rather ugly [(nested loops monster)](https://github.com/nofishonfriday/sws/blob/resources-settrackstatechunk-force-updating-fx-settings-set-fx-offline/SnM/SnM_Chunk.cpp#L953) but [seems to work](https://github.com/reaper-oss/sws/issues/1109#issuecomment-468943888).
As there are usually not 100s of FX on a track I hope it'll be ok (better alternatives always welcome :)).

`+` Clean up:
- REAPER API: make GetSetTrackGroupMembershipHigh(), TrackFX_GetOffline() mandatory
- SWS_IsTrackHeightLocked() tweaks/comment

`+` REAPER API: add TrackFX_SetOffline(), TakeFX_Get/SetOffline()
`+` ReaScript: add NF_TakeFX_GetModuleName()


nofishonfriday:
Closing because [fixed in REAPER](https://forum.cockos.com/showpost.php?p=2169479&postcount=1).

PR #1133, opened 2019-04-10
harden SNM_Get/SetIntConfigVar(), SNM_Get/SetDoubleConfigVar()

### LINKED ISSUE #1131: Accessing ConfigVar "g_config_project" crashes Reaper
Accessing the config-var "g_config_project" crashes Reaper or at least lets Lua hang up.
It is a crash within SWS, but as earlier versions worked without any problems, I think it is a regression of Reaper's side(if not, I'll report it into SWS-git as well).

Code:
```
  O=reaper.time_precise()

  A0=reaper.SNM_SetDoubleConfigVar("g_config_project", 100)
Tested on Windows 7, Reaper 5.974 and SWS 2.9.7 and SWS 2.10.0.1.
```

The rc2 of Reaper 5.974 seems to not have that problem, but I might be wrong.
5.973 and earlier worked without any problems.

According to Justin, this is due this variable being a string and should not be accessed as double.

The full post in this bug-reports-thread.
https://forum.cockos.com/showthread.php?p=2119876#post2119876

nofishonfriday:
fixes #1131

### DIFF
--- a/SnM/SnM_Misc.cpp
+++ b/SnM/SnM_Misc.cpp
@@ -204,14 +204,35 @@ bool SNM_GetProjectMarkerName(ReaProject* _proj, int _num, bool _isrgn, WDL_Fast
 	return false;
 }
 
+// https://forum.cockos.com/showthread.php?p=2119594#post2119594
+template <typename T>
+void* GetConfigVar_CheckSize(const char* cVar, T var)
+{
+	int sztmp;
+	void* p = NULL;
+	if (int iOffset = projectconfig_var_getoffs(cVar, &sztmp))
+	{
+		p = projectconfig_var_addr(EnumProjects(-1, NULL, 0), iOffset);
+	}
+	else
+	{
+		p = get_config_var(cVar, &sztmp);
+	}
+
+	if (p && (sztmp == sizeof(var)))
+		return p;
+	else
+		return NULL;
+}
+
 int SNM_GetIntConfigVar(const char* _varName, int _errVal) {
-	if (int* pVar = (int*)(GetConfigVar(_varName)))
+	if (int* pVar = (int*)(GetConfigVar_CheckSize(_varName, _errVal)))
 		return *pVar;
 	return _errVal;
 }
 
 bool SNM_SetIntConfigVar(const char* _varName, int _newVal) {
-	if (int* pVar = (int*)(GetConfigVar(_varName))) {
+	if (int* pVar = (int*)(GetConfigVar_CheckSize(_varName, _newVal))) {
 		*pVar = _newVal;
 		return true;
 	}
@@ -219,13 +240,13 @@ bool SNM_SetIntConfigVar(const char* _varName, int _newVal) {
 }
 
 double SNM_GetDoubleConfigVar(const char* _varName, double _errVal) {
-	if (double* pVar = (double*)(GetConfigVar(_varName)))
+	if (double* pVar = (double*)(GetConfigVar_CheckSize(_varName, _errVal)))
 		return *pVar;
 	return _errVal;
 }
 
 bool SNM_SetDoubleConfigVar(const char* _varName, double _newVal) {
-	if (double* pVar = (double*)(GetConfigVar(_varName))) {
+	if (double* pVar = (double*)(GetConfigVar_CheckSize(_varName, _newVal))) {
 		*pVar = _newVal;
 		return true;
 	}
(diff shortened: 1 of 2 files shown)


cfillion:
Why not  `sizeof(T)`? The `var` argument is not used at runtime and could be removed. It could also return a T* directly so that the function's users doesn't have to do the `static_cast` from void* every time. I think this would be simpler to use:

```cpp
double *value = GetConfigVar<double>("foobar");
```

Ideally, I think the original GetConfigVar should be completely replaced by the safer implementation, rather than adding another function. (If the checkless void* version is required somewhere else in the codebase for some reason, an explicit specialization could be made to accommodate that.)

Something like this:

```cpp
template <typename T>
T *GetConfigVar(const char *varName)
{
  int size = 0;
  void *value = NULL;
 
  if (const int offset = projectconfig_var_getoffs(varName, &size))
    value = projectconfig_var_addr(EnumProjects(-1, NULL, 0), offset);
  else
    value = get_config_var(varName, &size);

  if (size != sizeof(T))
    return NULL;
 
  return static_cast<T *>(value);
}
```


(The `SNM_GetXyzConfigVar` ReaScript functions could also be implemented as a single template if it wasn't for the APIFUNC macro, [reascript_vararg.php](https://gist.github.com/cfillion/fe2ffff40db5a198ebf085d2a4cc6de1), etc getting in the way... 😢  )

nofishonfriday:
closing temporarily to rework according to cfillion's suggestions

nofishonfriday:
@cfillion 
Thanks for the suggestions, appreciated (helps a lot to improve coding style. :))

> sizeof(var) could be sizeof(T)
I even had that at some point, but reverted for some reason, on retrospective, not sure why. :dagger: 

> Ideally...

[Done.](https://github.com/nofishonfriday/sws/commit/c67dc5df3de690dce21a5309b3c6cca0a0e4a4a5)
Seems to work, what do you think?

(I wasn't sure about the copyright note in configvar.h, if you want me to change/remove it, just tell...)

nofishonfriday:
@cfillion 
Can we do this to make the existing GetConfigVar() also use the new size check in configvar.h (to make existing calls safer, rather than rewriting all calls using the new function)  or is it a bad idea?

```
// sws.util.cpp
# include "configvar.h"

void* GetConfigVar(const char* cVar)
{
	return (void*)((uintptr_t)(GetConfigVar(cVar, NULL)));
}

cfillion:
I think all GetConfigVar/GetConfig/SetConfig should be hardened so that every configvars accesses in SWS are handled in the same way.

Here's another possible implementation, C++ OOP-style: [`harden-configvar-oop/configvar.h`](https://github.com/cfillion/sws/compare/cmake...cfillion:cmake+harden-configvar-oop#diff-ae25aaeee37ef4608b8aa1ff4fc349ebR29). I've search/replaced all GetConfigVar/GetConfig/SetConfig throughout the code. Get/SetConfig checked for whether the configvar was NULL, so I've preserved that (not sure if always necessary as SWS doesn't support REAPER < 5.965? I tried not to change existing behavior without thinking too much).

(I've also extracted a ConfigVarOverride utility class from one of my work-in-progress branches, it simplifies things like [this](https://github.com/cfillion/sws/compare/cmake...cfillion:cmake+harden-configvar-oop#diff-e7947e04de83d9b00015d5fb18e8c8b0L1765))

nofishonfriday:
Thanks.
I've seen your ConfigVarOverride utility already (yes, I sometimes sneak... :D), nice. Planned to you use this also in future, once merged.

So how do we proceed with this?
Seeing you incorporated this in your cmake branch I guess I leave my PR closed and you submit yours when done?

cfillion:
I based it off my cmake branch because it's easier for me to work with, but I can cherry-pick the commit into a fresh next-based branch before making any PR out if it.

The `explicit` keyword (used to prevent bug-inducing accidental conversion of ConfigVar to bool) is C++11 though. I doubt the very old version of clang currently used to make macOS builds knows about it.

I could of course temporarily remove it or wrap it around `__cplusplus` version checks. But, maybe it would be best to wait if and until C++11 is enabled before having that PR merged? (In which case, perhaps this PR should be reopened as a quick fix to #1131 in the meantime?)

nofishonfriday:
Ok, so I'll reopen it (for now).
If you have the new build system + conversion to C++ 11 ready before it gets merged, you can tell and I can close it again. :)

PR #1139, opened 2019-04-23
harden SNM_Get/SetIntConfigVar(), SNM_Get/SetDoubleConfigVar()

### LINKED ISSUE #1131: Accessing ConfigVar "g_config_project" crashes Reaper

nofishonfriday:
fixes #1131, Xcode project is also updated

This is a temporary fix.
@cfillion has a more comprehensive fix in the works, see https://github.com/reaper-oss/sws/pull/1133

### DIFF
--- a/SnM/SnM_Misc.cpp
+++ b/SnM/SnM_Misc.cpp
@@ -32,6 +32,7 @@
 #include "SnM_Misc.h"
 #include "SnM_Track.h"
 #include "SnM_Util.h"
+#include "../Utility/configvar.h"
 #ifdef _SNM_HOST_AW
 #include "../Misc/Adam.h"
 #endif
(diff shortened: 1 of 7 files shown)


nofishonfriday:
Closing because of @cfillion's more comprehensive PR #1191.

PR #1142, opened 2019-04-28
'Xenakios/SWS: Rename takes and source files...' tweaks

### LINKED ISSUE #1140: Rename Takes and Source Files leaves unused .reapeaks file and doubles filename extension
2 issues here:

#1 
I recently noticed that Rename Takes and Source Files leaves an unused .reapeaks file behind.

Say I have a file called "file.wav". That means on my HD I have 2 files
file.wav
file.wav.reapeaks

After renaming the file with Rename Takes and Source Files I have 3 files 
newname.wav
newname.wav.reapeaks
file.wav.reapeaks

#2
I also find little window which pops up for renaming a little misleading. Say I have a file called "file.wav" When I run Rename Takes and Source Files the window pops up and asks me to rename file.wav. When I type newname.wav the new file will be called "newname.wav.wav". It only works when I rename "file.wav" to "newname" (without the filename extension). Then it will rename it to newname.wav

nofishonfriday:
- automatically delete old .reapeaks files (also for Xenakios/SWS: Rename take source files...)
- don't double-append file extension if already contained in new filename/takename
- add (no undo) in action name

closes #1140

\+ S&M: Delete ... source file(s): remove redundant null check


Xenakios/TakeRenaming.cpp
```diff
@@ -161,6 +163,12 @@ void DoRenameSourceFileDialog666(COMMAND_T* ct)
 							{
 								GetSetMediaItemTakeInfo(alltakes[j],"P_SOURCE",newsrc);
 								delete thesrc;
+
+								// delete old .reapeaks file, #1140
+								char peakfn[SNM_MAX_PATH] = "";
+								GetPeakFileName(oldname.c_str(), peakfn, sizeof(peakfn));
+								if (peakfn && *peakfn != '\0')
```
cfillion: `peakfn` (pointer to char) is always non-null here (clang emits a [warning](https://travis-ci.com/cfillion/sws/jobs/196169940#L768) for this).

```suggestion
								if (*peakfn != '\0')
```

nofishonfriday: Caught [blindly copying code](https://github.com/reaper-oss/sws/blob/6963f7563851c8dd919db426bae825843939077f/SnM/SnM_Item.cpp#L1003)!

Changed, thanks.

nofishonfriday:
Made it a function (for reuseability).
Btw. cool to see the new auto build system in action. :)

PR #1193, opened 2019-09-10
sws-2019-08

nofishonfriday:
Putting it here for discussion/review.
This contains (almost, see below) all currently open PR's merged into next.
Builds here without errors (using VS2019/CMakeLists.txt)

Issues that would need to be discussed/reviewed:
1. PR #1111 vs. https://github.com/reaper-oss/sws/commit/b35a74484ca34f2a951e461ca76ef0a58aacbf66 :
These are conflicting commits, would need to decide if we continue to use chunking or revert to using API in BR_EnvelopeUtil.cpp. I originally merged PR #1111 (40220285b2520dbe9490c4181b15d35b8385087b) but later on decided to revert to using API because
- faster (though remembering cfillion said doesn't matter much)
- if at some point adding support for AIs we would need to use API anyway I think (as envelope chunks don't contain points in AIs currently)
I got some merge merge conflict in 33ce636d7865520e94657b73e2204cb0990c1f38 which I tried to resolve manually

2. I got quite a few merge conflict when merging PR #1191 which I also tried to resolve manually.
Attaching the files which Git produced showing the relevant locations here for reference.
[merging PR #1191 merge conflicts.zip](https://github.com/reaper-oss/sws/files/3597783/merging.PR.1191.merge.conflicts.zip)

@swstim, @cfillion
What do you think?


cfillion:
Looks like this contains commits from my `cmake` development branch. They should not be included. (The finalized `cmake-pr` branch that got merged into `next` had a different, cleaned up set of commits.)

I assume the merge conflicts from the configvar PR came from other PRs? It should merge cleanly with the current `next`.

nofishonfriday:
> Looks like this contains commits from my cmake development branch. They should not be included. > (The finalized cmake-pr branch that got merged into next had a different, cleaned up set of commits.)

> For instance the version.h include in cfillion.cpp was already fixed.

The version.h include in cfillion.cpp was a particular instance where I wasn't sure which of the conflicting code I should use. But other than that I'm rather sure I didn't merge anything from your cmake branch. Or do you see also other things from there?

> I assume the merge conflicts from the configvar PR came from other PRs? It should merge cleanly with the current next.

Yes, e.g. from
https://github.com/reaper-oss/sws/commit/3a720921d2e50aed087b98f5d284ea16c83bd4b8
and my PR ##1125

cfillion:
Ah, they came along with the cmake+harden-configvar-oop merge. The PR's (#1191) branch is just `harden-configvar-oop`.

nofishonfriday:
> Ah, they came along with the cmake+harden-configvar-oop merge. The PR's (#1191) branch is just harden-configvar-oop.

Duh, right, I see it now.
So should I discard the cmake+harden-configvar-oop and merge harden-configvar-oop instead?

Or what would you suggest should be next steps in general?
Actually I'm not even sure if this branch is supposed to be used for for next SWS version or the PRs get merged one by one into next?

PR #1197, opened 2019-09-13
add action SWS/NF: Toggle render speed (apply FX/render stems) realtime/not limited

### LINKED ISSUE #1065: FR: Toggle Action for Render Speed
Hi,

Thx to SWS we have 

- **Xenakios/SWS: Set render speed to not limited**
- **Xenakios/SWS: Set render speed to realtime**

Could we have a toggle action for this with on/off states so that it would be perfect in a toolbar ?

A scripts solution would involve having it launched a REAPER startup so that it can toggle the toolbar accordingly, not very handy :/

Thanks !

nofishonfriday:
rewrite of #1135 using new `configvar.h`
(needs #1191/#1195 merged to build)

closes #1065

### DIFF
--- a/nofish/nofish.cpp
+++ b/nofish/nofish.cpp
@@ -33,7 +33,8 @@
 #include "../Breeder/BR_ContinuousActions.h"
 #include "../Breeder/BR_Util.h"
 #include "../Breeder/BR_ReaScript.h" // BR_GetMouseCursorContext(), BR_ItemAtMouseCursor()
-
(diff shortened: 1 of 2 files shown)


pipelineaudio:
Since this is toggle, how would you know if its on or off? This looks like a great potential fix for Waves tune. Thank you!

cfillion:
The action reports its toggle state in the Action List. It should display as such in toolbars as well.

PR #1202, opened 2019-09-17
padreEnvelopeProcessor: fixes/tweaks

### LINKED ISSUE #1158: Bug: LFO Generator can't handle take envelopes with too much points:"Could not retrieve envelope object state!"
SWS Version 2.10.0 #1 built on Feb 6 2019

Having too much envelope points on a target item causes error: "Could not retrieve envelope object state!"
Getting to a point where there would be too much envelope points is allowed. Trying to come back from it is where there's a problem.

See video: https://youtu.be/o9Q716SJKyc

Doing the same testing with Track envelopes doesn't cause an error, but causes a slow hang when creating an extreme amount of envelope points, and when using the tool while an extreme amount exists.

Also, when using LFO Generator on a stretched item, results are very inconsistent.
Using action "Xenakios/SWS: Set item rate to 1.0 and pitch to 0.0" reveals that the LFO Generator is treating the item as if it was never stretched.

nofishonfriday:
take LFO generator:
- allow for generating more envelope points (use _GetEnvelopeStateChunkBig() instead of PadreGetEnvelopeState())
- account for take playrate
fixes #1158

padreUtils:
- remove GetEnvelopeState()

+Snapshots:
- TrackSends: use _GetEnvelopeStateChunkBig() (instead of GetEnvelopeStateChunk())

+add files: envelope.hpp/.cpp (place for contributor independent utility functions), update CMakeLists.txt

**rationale for adding  ~~UtilityFunctions.hpp/.cpp:~~** **envelope.hpp/.cpp**
Currently (general) utility functions are spread in various places in SWS:
sws_util / wol_Util / SnM_Util / XenUtils / ...

I thought I'd rather not add to this by adding my own NF_Util.
So in the past I've added functions to e.g. sws_util, but I don't consider it good style as these are not 'my' files.
UtilityFunctions.hpp/.cpp could serve as a place for collection of all future utility functions in one place (independent of contributor) and it also could be considered condensing some of the existing ones there. 

Wasn't sure about a good function naming convention there.
I've underscored (to differ from Reaper API functions) but not prefixed (e.g. with NF_) because of a possible contributor independent collection (see above), contributor name prefix seems not obligatory imo.

**Open to discussion/suggestions!** 

edit:
I see the builds failed. :(
Well, it does build here, need to investigate..

edit2:
https://travis-ci.com/reaper-oss/sws/jobs/236430802#L542
https://stackoverflow.com/questions/16861535/stdstring-has-no-member-named-front

Confused, I thought we're using C++11(+) now?
The other build errors I have solved I think.

### DIFF
--- a/ObjectState/TrackSends.cpp
+++ b/ObjectState/TrackSends.cpp
@@ -222,6 +222,34 @@ TrackSends::~TrackSends()
 	m_sends.Empty(true);
 }
 
+enum class EnvTypes {
+	AUXVOL = 0,
+	AUXPAN,
+	AUXMUTE
+};
+
+static void DisplayStoreSendEnvError(MediaTrack* tr, EnvTypes envType, const envelope::bad_get_env_chunk_big& ex)
+{
+	std::ostringstream ss;
+	ss << "Track " << static_cast<int>(GetMediaTrackInfo_Value(tr, "IP_TRACKNUMBER")) << ":\n";
+
+	switch (envType)
+	{
+	case EnvTypes::AUXVOL:
+		ss << __LOCALIZE("Error storing send volume envelope!\n", "sws_mbox");
+		break;
+	case EnvTypes::AUXPAN:
+		ss << __LOCALIZE("Error storing send pan envelope!\n", "sws_mbox");
+		break;
+	case EnvTypes::AUXMUTE:
+		ss << __LOCALIZE("Error storing send mute envelope!\n", "sws_mbox");
+		break;
+	}
+	ss << ex.what(); // localized in envelope.cpp
+
+	MessageBox(g_hwndParent, ss.str().c_str(), __LOCALIZE("SWS Snapshots - Error", "sws_mbox"), MB_OK);
+}
+
 void TrackSends::Build(MediaTrack* tr)
 {
 	// Get the HW sends from the track object string
@@ -270,30 +298,47 @@ void TrackSends::Build(MediaTrack* tr)
 			if (strncmp(line, searchStr, strlen(searchStr)) == 0) {
 	
 				// get the AUXVOL, AUXPAN, AUXMUTE subchunks
-				// note: chunks don't work properly with AIs currently
+				// note: envelope points in automations items aren't included in envelope chunks
 				// https://forum.cockos.com/showthread.php?t=205279
-				char AUXVOLchunk[4096]; char AUXPANchunk[4096]; char AUXMUTEchunk[4096];
-				WDL_FastString AUXVOLstr, AUXPANstr, AUXMUTEstr;
+				std::string AUXVOLstr, AUXPANstr, AUXMUTEstr;
 
 				TrackEnvelope* auxrcvVolEnv = (TrackEnvelope*)GetSetTrackSendInfo(pDest, -1, receiveIdx, "P_ENV", (void*)"<VOLENV");
 				TrackEnvelope* auxrcvPanEnv = (TrackEnvelope*)GetSetTrackSendInfo(pDest, -1, receiveIdx, "P_ENV", (void*)"<PANENV");
 				TrackEnvelope* auxrcvMuteEnv = (TrackEnvelope*)GetSetTrackSendInfo(pDest, -1, receiveIdx, "P_ENV", (void*)"<MUTEENV");
 
-				if (auxrcvVolEnv) {
-					GetEnvelopeStateChunk(auxrcvVolEnv, AUXVOLchunk, sizeof(AUXVOLchunk), true);
-					AUXVOLstr.Set(AUXVOLchunk);
+				if (auxrcvVolEnv) 
+				{
+					try {
+						AUXVOLstr = envelope::GetEnvelopeStateChunkBig(auxrcvVolEnv);
+					} 
+					catch (const envelope::bad_get_env_chunk_big &ex) {
+						
+						DisplayStoreSendEnvError(tr, EnvTypes::AUXVOL, ex);
+					}
 				}
 
-				if (auxrcvPanEnv) {
-					GetEnvelopeStateChunk(auxrcvPanEnv, AUXPANchunk, sizeof(AUXPANchunk), true);
-					AUXPANstr.Set(AUXPANchunk);
+				if (auxrcvPanEnv) 
+				{
+					try {
+						AUXPANstr = envelope::GetEnvelopeStateChunkBig(auxrcvPanEnv);
+					}
+					catch (const envelope::bad_get_env_chunk_big& ex) {
+						DisplayStoreSendEnvError(tr, EnvTypes::AUXPAN, ex);
+					}
+					
 				}
 
-				if (auxrcvMuteEnv) {
-					GetEnvelopeStateChunk(auxrcvMuteEnv, AUXMUTEchunk, sizeof(AUXMUTEchunk), true);
-					AUXMUTEstr.Set(AUXMUTEchunk);
+				if (auxrcvMuteEnv) 
+				{
+					try {
+						AUXMUTEstr = envelope::GetEnvelopeStateChunkBig(auxrcvMuteEnv);
+					}
+					catch (const envelope::bad_get_env_chunk_big& ex) {
+						DisplayStoreSendEnvError(tr, EnvTypes::AUXMUTE, ex);
+					}
+					
 				}
-				m_sends.Add(new TrackSend(&guid, line, AUXVOLstr.Get(), AUXPANstr.Get(), AUXMUTEstr.Get()));
+				m_sends.Add(new TrackSend(&guid, line, AUXVOLstr.c_str(), AUXPANstr.c_str(), AUXMUTEstr.c_str()));
 			}
 		}
 		SWS_FreeHeapPtr(trackStr);
--- a/Padre/padreEnvelopeProcessor.cpp
+++ b/Padre/padreEnvelopeProcessor.cpp
@@ -222,10 +222,13 @@ EnvelopeProcessor::ErrorCode EnvelopeProcessor::getTrackEnvelopeMinMax(TrackEnve
 	return eERRORCODE_OK;
 }
 
-void EnvelopeProcessor::writeLfoPoints(string &envState, double dStartTime, double dEndTime, double dValMin, double dValMax, LfoWaveParams &waveParams, double dPrecision, LfoWaveParams* freqModulator)
+void EnvelopeProcessor::writeLfoPoints(MediaItem_Take* take, string &envState, double dStartTime, double dEndTime, double dValMin, double dValMax, LfoWaveParams &waveParams, double dPrecision, LfoWaveParams* freqModulator)
 {
 	double dFreq, dDelay;
 	getFreqDelay(waveParams, dFreq, dDelay);
+
+	if (take) // account for take playrate, #1158
+		dFreq /= GetMediaItemTakeInfo_Value(take, "D_PLAYRATE");
 
 	double dMagnitude = waveParams.strength * (1.0 - fabs(waveParams.offset));
 	double dOff = 0.5*(dValMax+dValMin);
@@ -613,7 +616,7 @@ EnvelopeProcessor::ErrorCode EnvelopeProcessor::generateTrackLfo(TrackEnvelope* 
 		token = strtok(NULL, "\n");
 	}
 
-	writeLfoPoints(newState, dStartPos, dEndPos, dValMin, dValMax, waveParams, dPrecision);
+	writeLfoPoints(nullptr, newState, dStartPos, dEndPos, dValMin, dValMax, waveParams, dPrecision);
 
 	newState.append(token);
 	newState.append("\n");
@@ -748,7 +751,7 @@ EnvelopeProcessor::ErrorCode EnvelopeProcessor::generateTakeLfo(MediaItem_Take* 
 		token = strtok(NULL, "\n");
 	}
 
-	writeLfoPoints(newState, dStartPos, dEndPos, dValMin, dValMax, waveParams, dPrecision);
+	writeLfoPoints(take, newState, dStartPos, dEndPos, dValMin, dValMax, waveParams, dPrecision);
 
 	newState.append(token);
 	newState.append("\n");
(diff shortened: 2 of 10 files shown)


cfillion:
> `error: no member named 'front' in 'std::basic_string<char>'`
> Confused, I thought we're using C++11(+) now?

Ah, this is because we're targeting macOS 10.5. We can use C++11 syntax but we're ultimately limited to what's available in macOS 10.5's libc++ for any standard library features that require linking. Apple started to add support for C++11 in 10.7.

nofishonfriday:
Reading [Directly write into char* buffer of std::string](https://stackoverflow.com/questions/39200665/directly-write-into-char-buffer-of-stdstring/39200666):

so using `&str[0]` instead of `&str.front()` would be ok in this case as we have C++11 'environment'?

cfillion:
> Hm..if current standard library we're using doesn't comply to C++11 standard (basic_string object shall be stored contiguously) then probably not I think.

Right. The standard didn't guarantee it, but I'm pretty sure libstdc++ does (though it uses copy-on-write).

`operator[]`, `data` and `c_str` all give the same addresses here (tested on 10.6 and 10.10), implying they're all contiguous and null-terminated (since c_str always had to be).

```cpp
#include <string>
#include <iostream>

int main() {
  std::string a(1024, 'a');
  
  std::cout << (&a[0] == a.data()) << std::endl;  //=> true
  std::cout << (&a[0] == a.c_str()) << std::endl; //=> true
}
```

nofishonfriday:
Thanks!
Also found this which pretty much comes to the same conclusion:
https://stackoverflow.com/questions/760790/is-it-legal-to-write-to-stdstring

Phew, turned out more complicated (for me) than expected, but it finally builds now at least. :smile: 
If you'd get around to code review I'd appreciate.

ObjectState/TrackSends.cpp
```diff
@@ -270,30 +272,50 @@ void TrackSends::Build(MediaTrack* tr)
...
-					AUXVOLstr.Set(AUXVOLchunk);
+				if (auxrcvVolEnv) 
+				{
+					try {
+						AUXVOLstr = _GetEnvelopeStateChunkBig(auxrcvVolEnv);
+					} 
+					catch (_bad_get_env_chunk_big) {
+						
+						DisplayStoreSendEnvError(tr, "AUXVOL");
+						AUXVOLstr = "";
```
cfillion: `AUXVOLstr` is already initialized to an empty string by its constructor. There is no need re-assign it to a new empty value.

ObjectState/TrackSends.cpp
```diff
@@ -432,3 +454,25 @@ void TrackSends::GetChunk(WDL_FastString* chunk)
 	for (int i = 0; i < m_sends.GetSize(); i++)
 		m_sends.Get(i)->GetChunk(chunk);
 }
+
+void TrackSends::DisplayStoreSendEnvError(MediaTrack* tr, const char* envName)
```
cfillion: Instead of taking a const char* and doing string comparison, maybe this could take a Enum (integer comparison)?

ObjectState/TrackSends.cpp
```diff
@@ -432,3 +454,25 @@ void TrackSends::GetChunk(WDL_FastString* chunk)
...
+		msg.append(__LOCALIZE("Error storing send volume envelope!", "sws_mbox"));
+	}
+	else if (strncmp(envName, "AUXPAN", 6) == 0) {
+		msg.append(__LOCALIZE("Error storing send pan envelope!", "sws_mbox"));
+	}
+	else if (strncmp(envName, "AUXMUTE", 7) == 0) {
+		msg.append(__LOCALIZE("Error storing send mute envelope!", "sws_mbox"));
+	}
+
+	MessageBox(g_hwndParent, msg.c_str(), __LOCALIZE("SWS Snapshots - Error", "sws_mbox"), MB_OK);
```
cfillion: The message could be built using only the stringstream.

```cpp
std::ostringstream ss;
ss << "Track " << static_cast<int>(GetMediaTrackInfo_Value(tr, "IP_TRACKNUMBER")) << ":\n"

switch(envType) {
case AUXVOL:
  ss << __LOCALIZE("Error storing send volume envelope!", "sws_mbox");
  break;
case AUXPAN:
  ss << __LOCALIZE("Error storing send pan envelope!", "sws_mbox");
  break;
// ...
}

MessageBox(..., ss.str().c_str(), ...);

File: Padre/padreEnvelopeProcessor.cpp [hunk 4]
```diff
@@ -222,11 +223,18 @@ EnvelopeProcessor::ErrorCode EnvelopeProcessor::getTrackEnvelopeMinMax(TrackEnve
...
 	double dFreq, dDelay;
 	getFreqDelay(waveParams, dFreq, dDelay);
 
+	// account for take playrate, #1158
+	if (take)
+	{
+		double playrate = GetMediaItemTakeInfo_Value(take, "D_PLAYRATE");
+		dFreq = dFreq / playrate;
+	}
+
```
cfillion: This seems unrelated to the buffer change, and should probably be in a separate commit (doesn't have to be a new branch and PR, just a commit)?

Utility/UtilityFunctions.cpp
```diff
@@ -0,0 +1,73 @@
...
+******************************************************************************/
+
+#include "stdafx.h"
+#include "UtilityFunctions.hpp"
+
+// wraps GetEnvelopeStateChunk() to make it more save
+// throws _bad_get_env_chunk_big on failure
+// until maybe it gets enhanced one day
+// https://forum.cockos.com/showthread.php?p=2142245#post2142245
+std::string _GetEnvelopeStateChunkBig(TrackEnvelope* envelope, bool isUndo /*= false*/)
```
cfillion: In my opinion this should just be `GetEnvelopeStateChunkBig`. The leading underscore makes it look like it's some kind of private/special/secret function that shouldn't be used.

If there is a risk of name conflict, maybe namespacing would be a better option here.

Utility/UtilityFunctions.cpp
```diff
@@ -0,0 +1,73 @@
...
+{
+	_bad_get_env_chunk_big exExceeding100MiB("_GetEnvelopeStateChunkBig() exception (>  100 MiB)");
+	_bad_get_env_chunk_big exBad_alloc("_GetEnvelopeStateChunkBig() exception (bad_alloc)");
```
cfillion: There is no need to construct the exception object ahead of time. `throw bad_get_env_chunk_big("Reason")` should be enough.

Utility/UtilityFunctions.cpp
```diff
@@ -0,0 +1,73 @@
...
+		const size_t endpos = buffer.find('\0');
+
+		if (endpos != std::string::npos) {
+			buffer.resize(endpos);
+			return buffer;
+		}
+
+		if (buffer.size() > 100 << 20) { // 100 MiB
+			throw exExceeding100MiB;
+			std::cout << "_GetEnvelopeStateChunkBig() exception (>  100 MiB)";
```
cfillion: Leftover from testing?

Utility/UtilityFunctions.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#pragma once
+
+std::string _GetEnvelopeStateChunkBig(TrackEnvelope* envelope, bool isUndo = false);
+
+class _bad_get_env_chunk_big : public std::exception {
+private:
+	std::string msg;
+public:
+	explicit _bad_get_env_chunk_big(const std::string& message) {};
```
cfillion: The message is lost here, what() will always return an empty string. You could make use of std::runtime_error's constructor instead of re-writing it:

```cpp
class bad_get_env_chunk_big : public std::runtime_error {
public:
  using runtime_error::runtime_error;
};
```

Utility/UtilityFunctions.cpp
```diff
@@ -0,0 +1,73 @@
...
+/ NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR COPYRIGHT
+/ HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY,
+/ WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+/ FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR
+/ OTHER DEALINGS IN THE SOFTWARE.
+/
+******************************************************************************/
+
+#include "stdafx.h"
+#include "UtilityFunctions.hpp"
```
cfillion: "UtilityFunctions" is rather vague. How about a new contributor-independent envelope.cpp?

ObjectState/TrackSends.cpp
```diff
@@ -29,6 +29,8 @@
 #include "../reaper/localize.h"
 #include "../SnM/SnM_Dlg.h"
 #include "TrackSends.h"
+#include "Utility/UtilityFunctions.hpp"
+#include <sstream>
```
cfillion: The include could be put in the precompiled header `stdafx.h` (it's only precompiled on Windows).

nofishonfriday:
Thanks for the comprehensive review!

```
class bad_get_env_chunk_big : public std::runtime_error {
public:
  using runtime_error::runtime_error;
};
```

Hm..fine on Win/Linux, no luck on Mac:
https://travis-ci.com/reaper-oss/sws/jobs/242268219#L523

Any suggestion?

cfillion:
Adding `#include <stdexcept>` should do the trick.

nofishonfriday:
Thanks.

> Also, maybe this could take the exception (const bad_get_env_chunk_big &ex) and display its message (ex.what()) to further explain what the error is about (100 MiB buffer limit reached vs memory allocation failed).

I had some difficulties with this:
- ```bad_get_env_chunk_big``` is imported in ```padreEnvelopeProcessor.cpp```, so I couldn't use it in the function definition in ```padreEnvelopeProcessor.h```

Now takes the exception string instead

- ```__LOCALIZE(exception.c_str(), "sws_mbox");``` doesn't seem to work (I wonder why, as it does seem to take a const char[]?)

Used ```__localizeFunc()``` instead, is this ok?
https://github.com/reaper-oss/sws/commit/1460bdf517307558e6d6a79d29d7580bd095218d#diff-cd23b1cfff5934876e132daa9092852bR471

cfillion:
> `__LOCALIZE(exception.c_str(), "sws_mbox");` doesn't seem to work (I wonder why, as it does seem to take a const char[]?)

I'll look into it more closely tomorrow (I'll have to check whether the langpack generation tool would detect __localizeFunc). Alternatively `__LOCALIZE` could be used when the exception is being thrown.

> bad_get_env_chunk_big is imported in padreEnvelopeProcessor.cpp, so I couldn't use it in the function definition in padreEnvelopeProcessor.h

You should be able to use `bad_get_env_chunk_big` as a function parameter in TrackSends.h without including envelope.hpp by [forward-declaring](https://en.cppreference.com/w/cpp/language/class#Forward_declaration) it:

```cpp
namespace sws {
  class bad_get_env_chunk_big;
};

void DisplayStoreSendEnvError(MediaTrack* tr, EnvTypes envName, const envUtil::bad_get_env_chunk_big &);
```

However, because DisplayStoreSendEnvError does not use any members of the TrackSends class and is not used outside of the TrackSends.cpp, it does not need to be put in the header at all. It could be a static function living only within the scope of the .cpp.

```cpp
static void DisplayStoreSendEnvError(MediaTrack* tr, EnvTypes envType, const envUtil::bad_get_env_chunk_big &ex)
{
}

// code that uses it below (or add a prototype for the function before said code)
```

nofishonfriday:
> However, because DisplayStoreSendEnvError does not use any members of the TrackSends class and is not used outside of the TrackSends.cpp, it does not need to be put in the header at all. It could be a static function living only within the scope of the .cpp.

The only thought to declare it in the header was avoiding using forward declarations / having to put the function somewhere in the middle (both  I found a bit cluttery) ;D
But I see your point, will do. :)

nofishonfriday:
> It could be a static function living only within the scope of the .cpp.

Done.
Also moved `enum EnvTypes` from `TrackSends.h` to `TrackSends.cpp` (making it strongly typed enum, default to int, C++11).

> Alternatively __LOCALIZE could be used when the exception is being thrown.

Done.

Also changed `namespace envUtil` to `namespace envelope` (actually I've not yet seen a camel-cased namespace :D)
Maybe could even be `namespace env` for less typing (similar to std::)?

Btw. why did you suggest `envelope.hpp/cpp` - lowercase?
Somehow I have in mind filenames always capitalized, any rule you follow here?

Utility/envelope.cpp
```diff
@@ -0,0 +1,71 @@
...
+#include "../reaper/localize.h"
+#include "envelope.hpp"
+
+namespace envelope
+{
+	// wraps GetEnvelopeStateChunk() to make it more save
+	// until maybe it gets enhanced one day
+	// https://forum.cockos.com/showthread.php?p=2142245#post2142245
+	// throws envUtil::bad_get_env_chunk_big on failure
+	std::string GetEnvelopeStateChunkBig(TrackEnvelope* envelope, bool isUndo /*= false*/)
```
cfillion: You can save an indentation level in this file using `std::string envelope::GetEnvelopeStateChunkBig` in this line and removing the namespace (and no need to repeat the namespace name within the function).

nofishonfriday: But then every further declared function would have to use envelope:: also, no?
Is the repetitiveness worth saving the indentation level you think? :)

cfillion: It's the same as when defining class member functions.

nofishonfriday: Point taken, I'll remove the namespace { }.

nofishonfriday:
done https://github.com/reaper-oss/sws/pull/1202#pullrequestreview-297791857

Utility/envelope.cpp
```diff
@@ -0,0 +1,69 @@
...
+		const size_t endpos = buffer.find('\0');
+
+		if (endpos != std::string::npos) {
+			buffer.resize(endpos);
+			return buffer;
+		}
+
+		if (buffer.size() > 100 << 20) { // 100 MiB
+			throw bad_get_env_chunk_big(__LOCALIZE("The envelope chunk size exceeded the 100 MiB limit.", "sws_mbox"));
+			break;
```
cfillion:
```suggestion
```

This `break` is never reached (like the previous `std::cout`): once the exception is thrown the program execution jumps right away to the first matching catch block found in the stack (or aborts the program's execution if it's not being catched).


I have no more complaints after this. 😄

nofishonfriday: Ah, iirc that's a leftover I overlooked from your original code from Slack I think (where there wasn't any `try - catch`), thanks.

If I knew this gets so involved I wouldn't have assigned. :D
Just kidding, great learning experience once again.
Thanks again for sticking with me. :)

PR #1217, opened 2019-10-29
"SWS/S&M: Select FX x for selected tracks": use  TrackFX_Show() instead of chunk patching if possible

nofishonfriday:
I've done this some time ago but wanted to wait with submitting until the big SWS rework was finished.

### DIFF
--- a/SnM/SnM_FX.cpp
+++ b/SnM/SnM_FX.cpp
@@ -344,25 +344,27 @@ void UpdateAllFXsBypassSelItems(COMMAND_T* _ct) {
 // Track FX selection
 ///////////////////////////////////////////////////////////////////////////////
 
-int SelectTrackFX(MediaTrack* _tr, int _fx)
+bool SelectTrackFX(MediaTrack* _tr, int _fx)
 {
-	int updates = 0;
-	if (_tr && _fx >=0 && _fx < TrackFX_GetCount(_tr))
+	if (_tr && _fx >= 0 && _fx < TrackFX_GetCount(_tr))
 	{
-		SNM_ChunkParserPatcher p(_tr);
-		char pLastSel[4]="", pShow[4]=""; // 4 should be enough..
-		if (snprintfStrict(pLastSel, sizeof(pLastSel), "%d", _fx) > 0)
+		if (TrackFX_GetChainVisible(_tr) != -1) // FX chain is open, can use API
 		{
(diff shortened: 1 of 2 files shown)


cfillion:
Can you move the whatsnew line to the Miscellanious section below the Quantize one (so that it's alphabetically sorted)? (Unless you put it at the top to wait for the next release cycle?)

Also I'm thinking something along the line of `+Improve performance of "SWS/S&M: Select FX x for selected tracks" when the FX chain is opened` would have more meaning for those who don't know what API or chunk parsing is.

Additional improvement ideas:
- Looks like `updated` is only written to and used as a boolean (1/0). If so, it should be one (also in the other overload of `SelectTrackFX` where the return value is used).
- `pLastSel` is only used by `ParsePatch` and can be moved to its `else` block.

nofishonfriday:
> Unless you put it at the top to wait for the next release cycle?
That was the thought. :) (Thinking it's not a critical bugfix and this release has a lots of other new stuff already.)
Will do the other suggestions, thanks.

nofishonfriday:
updated.

cfillion:
Found a bug (not new): the undo point is only created if the last selected track is updated.

`updated =` should be `updated ||=`

nofishonfriday:
Oh so it **did** turn into a bugfix eventually. :D
I'll add the whatsnew line within v2.11.0 section then, so it could be added to the release.

nofishonfriday:
Thanks.

> I did a search for updated = [^tf] and found a few suspicious matches. I'll check (and possibly fix) those too.

Sory, could you rephrase in layman's terms please? :P
Is it about line endings, if there's anything I can enhance on my side let me know. :)

nofishonfriday:
Getting what you're doing but gotta say the line seen in your pic `ag 'updated = [^tf]'` leaves to totally stumped.
If you're in the modd I'd be happy if you dissect it for me, otherwise just ignore. :)

cfillion:
It is a regular expression: `[^tf]` is a "character class" that matches any character that is not `t` nor `f`. The intent is to filter out `updated = true` and `updated = false` from the results.

(That being said a better–but longer to type–regex would have been `updated = (?!true|false)` where `(?!...)` is a "negative lookahead". This wouldn't filter out expressions such as `updated = toto_bunny` just because it begins with a `t`.)

PR #1229, opened 2019-12-03
Loudness: add dual mono scanning mode [for after v2.12.0]

### LINKED ISSUE #1209: FR: Loudness - add dual mono scanning mode (was: Bug in SWS - problem detected about loudness)
Hello, I use SWS a lot, they are very useful to me. But I detected a measurement problem with loudness

It turns out that a while ago I found a contradiction in the reading of the integrated between Youlean and what the SWS measured. There were exactly 3LU of difference

I made measurements with white noise clips. And testing. I found where the error was

Apparently: it does not measure items that are "mono" correctly. With stereo, there is no problem. But with the "mono", there is 3dB difference.

The sound intensity result is even perceptible to the listening, so here I attach a project where you can make the measurements, compare readings between Youlean and SWS, and draw your own conclusions from this error.

https://mega.nz/#!SkAFTI7A!lXn_a8XFYLOO9BK0e2l18KT3mxn9prR2CqseDZL4vLU

1st Item: white noise, stereo in -8LUFS for both Youlean and SWS
2nd Item: white noise, mono at -5 for Youlean and -8 for SWS (bug)
3rd Item: white noise, stereo in -5 for Youlean and for SWS

This not only happens with white noise, it also happens to me with mono recordings that I make, there is always that bug present.

Conclusion: LUFS reading from SWS is failing for mono items. Even using "High precision mode" 

thank you

nofishonfriday:
closes #1209


nofishonfriday:
reminder to self: should disable the unused channels here:
https://github.com/reaper-oss/sws/pull/1229/files#diff-ec4419167b2eded4f3b075cf8db5e183R1080

(Will do if PR #1232 is merged.)

PR #1230, opened 2019-12-05
add action: SWS/NF: Force media online (all takes)

nofishonfriday:
As there seems [no native action](https://forum.cockos.com/showthread.php?t=227996) to do this currently, it would come in handy for a project I'm working on, so it'd be nice if it could make it in for v.2..11.0 official. :)

nitpicky thought:
I've removed the _Add_ in the _New actions:_ section in` whatsnew.txt` as it seems kinda redundant?

### DIFF
--- a/nofish/nofish.cpp
+++ b/nofish/nofish.cpp
@@ -30,10 +30,15 @@
 #include "stdafx.h"
 #include "nofish.h"
 
+#include <regex>
+
 #include "../Breeder/BR_ContinuousActions.h"
-#include "../Breeder/BR_Util.h"
 #include "../Breeder/BR_ReaScript.h" // BR_GetMouseCursorContext(), BR_ItemAtMouseCursor()
+#include "../Breeder/BR_Util.h"
+#include "../SnM/SnM_Chunk.h" // SNM_ChunkParserPatcher
+#include "../sws_util.h" // SWS_GetSelectedMediaItems()
 #include "../Utility/configvar.h"
+
 
 //////////////////////////////////////////////////////////////////
 //                                                              //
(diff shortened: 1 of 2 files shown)


cfillion:
C++11 regexes are not available on macOS 10.5.

From the Travis log:
```cpp
/Users/travis/build/reaper-oss/sws/nofish/nofish.cpp:33:10: fatal error: 'regex' file not found
#include <regex>
         ^~~~~~~
1 error generated.
make[2]: *** [CMakeFiles/sws.dir/nofish/nofish.cpp.o] Error 1
```

Is there some advantage to doing this action in C++ rather than as a script? Speaking of which looks like `SWS_GetSelectedMediaItems` could be optimized using `CountSelectedMediaItems` (pre-resize the buffer) and `GetSelectedMediaItem` (no need to loop over tracks).

Here's an regex-less implementation (not using std::string to avoid copying when possible). 
This won't work if the chunk happens to contain _OFFLINE_ elsewhere though (eg. take name, item notes, source filename...).

```cpp
for (int i = 0; i < items.GetSize(); i++)
{
	MediaItem* item = items.Get()[i];
	SNM_ChunkParserPatcher p(item);
	p.SetWantsMinimalState(true);

	WDL_FastString *chunk = p.GetChunk();
	if (char *match = strstr(chunk->Get(), "_OFFLINE_")) {
		chunk->DeleteSub(match - chunk->Get(), strlen("_OFFLINE_"));
		p.IncUpdates();
	}
}
```

nofishonfriday:
@cfillion 
Thanks.
>Is there some advantage to doing this action in C++ rather than as a script?

I don't think so. :) Only reason I did it in SWS is it was the easiest for me to do (don't know anything about Lua string processing) and I needed a working solution quickly because I'm in the middle of the deadlined project I need it for (*red face*).
Script solution would be totally fine otherwise though.

nofishonfriday:
Hm..since I don't have capacity to work on this further currently unfortunately, probably best is closing this PR (for now, or maybe try script version later) and keep on using the current version privately until this project is finished (mid December).

PR #1258, opened 2019-12-17
Added and fixed various grid actions

Breeder:
Added SWS/AW arrange grid actions (Issue 1244)
Added SWS/NF action to toggle swing grid in MIDI editor (Issue 1243)

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -45,10 +45,11 @@ for:
 
   - matrix: { only: [ APPVEYOR_BUILD_WORKER_IMAGE: &windows Visual Studio 2019 ] }
     install:
-      - choco install -y nsis php
+      - '"%ProgramFiles(x86)%\NSIS\uninst-nsis.exe" /S'
+      - choco install -y nsis-unicode php
       - refreshenv
       - curl -fsSLO https://nsis.sourceforge.io/mediawiki/images/1/18/NsProcess.zip
-      - 7z e nsProcess.zip -o"%ProgramFiles(x86)%\NSIS\Plugins\x86-unicode" nsProcessW.dll -r
+      - 7z e nsProcess.zip -o"%ProgramFiles(x86)%\NSIS\Unicode\Plugins" nsProcessW.dll -r
       - if "%ARCH%" == "x64" call "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars64.bat"
       - if "%ARCH%" == "x86" call "%ProgramFiles(x86)%\Microsoft Visual Studio\2019\Community\VC\Auxiliary\Build\vcvars32.bat"
     build_script:
--- a/setup/win32/NSIS.template.in
+++ b/setup/win32/NSIS.template.in
@@ -1,6 +1,9 @@
 !include MUI2.nsh
+!include Sections.nsh
 
-Unicode true
+!ifndef NSIS_UNICODE
+  Unicode true
+!endif
 
 AllowSkipFiles off
 Name "@CPACK_NSIS_PACKAGE_NAME@"


Breeder/BR.cpp
```diff
@@ -44,17 +44,35 @@
...
+static COMMAND_T* g_nextActionApplyer   = NULL;
+static int        g_nextActionToApply   = 0;
+static int        g_nextActionLoCmd     = 0;
+static int        g_nextActionHiCmd     = 0;
 static set<int>   g_nextActionApplyers;
 
 /******************************************************************************
 * Command hook                                                                *
 ******************************************************************************/
+static void DeferRefreshToolbar()
```
cfillion: Is there a reason for waiting two timer cycles before refreshing the toolbars? I'd expect a function like this to remove the timer at the next cycle without an extra flag to toggle and check.

Misc/Adam.cpp
```diff
@@ -2673,47 +2674,176 @@ int IsGridDotted(COMMAND_T* = NULL)
...
 	if (IsGridTriplet())
 		*pGridDiv *= 3.0/2.0;
 	else
 		*pGridDiv *= 2.0/3.0;
 
-
+	if (MIDIEditor_GetActive() && GetToggleCommandStateEx(SECTION_MIDI_EDITOR, 41022))     // Grid: Use the same grid division in arrange view and MIDI editor
+		if (IsGridTriplet() && !GetToggleCommandStateEx(SECTION_MIDI_EDITOR, 41004))       // Grid: Set grid type to triplet
+			MIDIEditor_LastFocused_OnCommand(41004, false);                                // Grid: Set grid type to triplet
+		else if (!IsGridTriplet() && !GetToggleCommandStateEx(SECTION_MIDI_EDITOR, 41003)) // Grid: Set grid type to straight
```
cfillion: While this code does behave as the indentation leads to believe, braces should be added around the first `if` to fix these warnings from Clang:

```
../Misc/Adam.cpp:2732:3: warning: add explicit braces to avoid dangling else [-Wdangling-else]
                else if (!IsGridTriplet() && !GetToggleCommandStateEx(SECTION_MIDI_EDITOR, 41003)) // Grid: Set grid type to straight
                ^
../Misc/Adam.cpp:2755:3: warning: add explicit braces to avoid dangling else [-Wdangling-else]
                else if (!IsGridDotted() && !GetToggleCommandStateEx(SECTION_MIDI_EDITOR, 41003)) // Grid: Set grid type to straight
                ^
2 warnings generated.
```

Misc/Adam.cpp
```diff
@@ -2736,14 +2866,34 @@ void AWToggleClickTrack(COMMAND_T*)
...
+		NamedCommandLookup("_SWS_SETGRID_PRESERVE_TYPE_1_2"),
+		NamedCommandLookup("_SWS_SETGRID_PRESERVE_TYPE_1_4"),
+		NamedCommandLookup("_SWS_SETGRID_PRESERVE_TYPE_1_8"),
+		NamedCommandLookup("_SWS_SETGRID_PRESERVE_TYPE_1_16"),
+		NamedCommandLookup("_SWS_SETGRID_PRESERVE_TYPE_1_32"),
+		NamedCommandLookup("_SWS_SETGRID_PRESERVE_TYPE_1_64"),
+		NamedCommandLookup("_SWS_SETGRID_PRESERVE_TYPE_1_128")
 	};
-	for (int i = 0; i < 3; ++i)
+	for (int i = 0; i < 14; ++i)
```
cfillion: With C++11 now being enabled, range-based for loops can be used. Hard-coding the size of the array can be completely avoided:
```cpp
static const int cmds[] {
	NamedCommandLookup("_SWS_AWTOGGLETRIPLET"),
	NamedCommandLookup("_SWS_AWTOGGLEDOTTED"),
	// ...
};
```

```suggestion
	for (const int cmd : cmds)
```

nofish/nofish.cpp
```diff
@@ -80,14 +81,24 @@ int Main_IsMIDIGridDotted(COMMAND_T* = NULL)
 	return toggleState;
 }
 
+int Main_IsMIDIGridSwing(COMMAND_T* = NULL)
```
cfillion: This can be simplified a little:

```cpp
// Grid: Set grid type to straight/triplet/dotted
for (int cmd = 41003; cmd <= 41005; ++cmd) {
	if (GetToggleCommandStateEx(ME_SECTION, cmd))
		return 0;
}

return 1;
```

Misc/Adam.cpp
```diff
@@ -2673,47 +2674,176 @@ int IsGridDotted(COMMAND_T* = NULL)
...
+	else if (IsGridDotted())
+	{
+		if (*pGridDiv == normalizedGrid * (3.0 / 2.0))
+			toggleState = 1;
+	}
+	else
+	{
+		if (*pGridDiv == normalizedGrid)
+			toggleState = 1;
+	}
```
cfillion: This can be simplified since all branches are doing the same thing:

```cpp
if (IsGridTriplet())
	normalizedGrid *= 2.0 / 3.0;
else if (IsGridDotted())
	normalizedGrid *= 3.0 / 2.0;

return *pGridDiv == normalizedGrid;
```

Breeder/BR.cpp
Breeder: Yes, it is needed because it needs to refresh AFTER MIDI editor opens...

Misc/Adam.cpp
Breeder: So add braces then :)

Misc/Adam.cpp
Breeder: Ah nice to know SWS is full C++11 now :) Thank you

nofish/nofish.cpp
Breeder: Some MIDI editor actions for grid report state and some do not, it's best to leave it verbose to know what's going on.

Misc/Adam.cpp
Breeder: This is testing for the toggle state, not setting grid division. I really recommend you leave everything I did as I did it because I invested a lot of times into testing it and making sure everything gets properly refreshed everywhere.

Misc/Adam.cpp
Breeder: Btw, I like to fix warnings too ;) But I don't remember seeing that one!
But for sake of readability, I always skip braces when I can...it's just my style

Breeder/BR.cpp
cfillion: Isn't the MIDI editor being opened synchronously when the action 40153 is run? (so always before the next defer cycle)

I added a breakpoint in `DeferRefreshToolbar` and it's always hit after the ME has opened. Also `MIDIEditor_GetActive` returns non-null the first time `DeferRefreshToolbar` is called.

Misc/Adam.cpp
cfillion: You can find Clang's warnings in the Travis build log for this PR: https://travis-ci.com/reaper-oss/sws/jobs/268161882#L460.

I'd prefer it if you'd amend the commit and force-pushed the branch yourself. I can do it if you want me to (as long as "Allow edits from maintainers" is enabled to the right of this page), but your local branch will diverge.

Misc/Adam.cpp
cfillion: Note that only syntax features are enabled. For compatibility with macOS 10.5, the C++11 standard library is not available.

nofish/nofish.cpp
cfillion: I see, good reason. Before I noticed all three `GetToggleCommandStateEx` could be merged into one, my original idea was to suggest this equivalent code to avoid doing `if(1) return 1; else return 0;`:

```cpp
return GetToggleCommandStateEx(ME_SECTION, 41005) == 0 && // Grid: Set grid type to dotted
       GetToggleCommandStateEx(ME_SECTION, 41004) == 0 && // Grid: Set grid type to triplet
       GetToggleCommandStateEx(ME_SECTION, 41003) == 0;   // Grid: Set grid type to straight
```

(...or alternatively, as you mentioned preferring to skip braces when possible–which I also do, each branch is 1 line so the braces can be removed here.)

Misc/Adam.cpp
cfillion: The refactored code does exactly the same thing as the original. It does not set the grid division either.

These are only improvement suggestions though, I'm not going to edit someone else's PR just like that. 😉

Breeder/BR.cpp
Breeder: When I tested it, it didn't work until I added this.

Misc/Adam.cpp
Breeder: I just look at VS warnings, but it had some many warnings I didn't catch this one.

Breeder: I'll enable it

Misc/Adam.cpp
Breeder: Yikes, this is why I always stuck with old C++ ways, cause I wasn't able to test directly on OSX

nofish/nofish.cpp
Breeder: I really don't mind if someone knowledge updates my code, so please, never hesitate to do so!

Misc/Adam.cpp
Breeder: You can always edit my stuff man, this is open-source and we're all contributing somehow. I know you know your stuff so I really don't mind. Just please make sure to test any changes, I really had unexpected amount of work to have all of this working properly due to all "refresh" things because you gotta refresh both the arrange and MIDI editor section

Misc/Adam.cpp
Breeder: It was already enabled ;D

cfillion:
I fixed the Clang warnings and did some of the cleanups discussed above. I tested on REAPER v5 and v6, macOS and Windows.

I also found (and fixed) a bug in the new "Set grid to X preserving grid type" actions: they were disabling swing if the grid was set to a triplet.

> When I tested it, it didn't work until I added this.

I could not reproduce that. Here it works in the very next defer cycle without having to wait for another one. Could you double-check the builds below (once they're posted) to be 100% sure?

Breeder:
THIS IS NOT A BUG! They were disabling swing intentionally because you can either have swing, triplet or dotted. Same as midi editor and same as any musical theory states. How can you have triplet swing? Swing swings the second note of a two note part piece, in case of triple what it would swing? The second or the third? Please, revert this as it makes no sense to have both swing and triplet enabled at the same time.

Breeder:
And regarding that timer fix, please leave it be because it worked like that on my configuration and that one timer thing during opening of the MIDI editor won't harm anyone and it won't affect performance.

You are wasting your time trying to "fix" this because it isn't broken.

cfillion:
> How can you have triplet swing? What it would swing? The second or the third?

It's very weird (swings every two notes making the next triplet not line up with the beat), yet it is a combination REAPER allows in both the arrange view and in the MIDI editor. https://i.imgur.com/IdwGpMh.png

I wouldn't be surprised if someone had a use for that... Anyway, I will remove my erroneous fix commit (I separated that change from the others for easy reversal in the case the original "strict" behavior was indeed preferred).

> Maybe it's settings, maybe it's you using dev builds (just guessing here) and me standard builds - I have no clue but I know that during testing, MIDI editor toolbar buttons didn't refresh until the second iteration and by disabling this you will just get bug reports in the future. 

Mhh that's interesting. Can you confirm the latest build above don't work on your system? How about in a fresh portable install?

I don't mind restoring the "wait one extra cycle" hack (will do if I don't find anything after doing more testing), but without any understanding of the underlying reason it feels very fragile.

If there is some sort of non-understood timing variation, it is possible for some users it would be more than two timer cycles, and we'd get bug reports in the future anyway. I think it's worth investigating why and what happens when it doesn't work to be sure the toolbars are always correctly refreshed. 

For example, perhaps waiting until the MIDI editor is actually open before refreshing the toolbars and stopping the timer would be a more robust solution? (Assuming that–or something similar that can be checked at runtime–turns out to be the root cause of the issue)

(I tested using 5.99 and 6.02, the later being a fresh portable install, debug and release sws builds, all 64-bit.)

> I feel I'm being abused here to be honest...doubted and called out for minuscule things like the way I'm writing braces

I am very sorry you feel this way. My intent is not to doubt or call out, but rather to do a proper peer code review (as is common in a lot of teams, useful and considered a good development practice).

(Of course I fully expect to receive and welcome code reviews on my own contributions as well. I want to be told if I let a bug through, didn't think of a possible better implementation, or just plain forgot to put a space between the `if` and its following opening parenthesis.)

(The braces thing was to fix Clang's `dangling-else` warning. SWS builds without warnings now–excluding third-party code–so it would have been a shame to introduce two warnings over mere braces. I don't see a reason to disable `dangling-else` either.)

cfillion:
I restored the wait-two-timer-cycle hack and removed the erroneous support for REAPER's strange swig triplets so this PR can be included in v2.12.

Did you have a chance to check whether the toolbar refresh issue you were having was caused by *Extensions > SWS Options > Enable toolbars auto refresh* being disabled?

`_NF_ME_TOGGLETRIPLET`, `_NF_ME_TOGGLEDOTTED` and `_NF_ME_TOGGLESWING` are continuously refreshed every 300 ms with the default settings.

If that option was indeed the culprit, I think `DeferRefreshToolbar` should be removed. Wouldn't automatically doing a toolbar refresh when opening the MIDI editor while the option to do so was explicitely disabled by the user be undesirable?

cfillion:
Oops, GitHub automatically closed this as I pushed the wrong branch (I did `git push -f breeder master` instead of `git push -f breeder breeder-grid-actions:master`... d'oh). I'll open a new PR with your original commit intact + my corrected cleanup/warning fix commit unless you'd prefer otherwise.

https://github.com/reaper-oss/sws/compare/master...cfillion:breeder-grid-actions

I'll wait a little just in case `DeferRefreshToolbar` can indeed be safely removed (as explained above).

nofishonfriday:
It seems I can produce a crash if
- assign _SWS/AW: Toggle triplet grid_ and _SWS/AW: Toggle swing grid_ to toolbar buttons
- enable both via Reaper
- turn triplet grid off off via button press

https://i.imgur.com/SFWQVqt.gif

cfillion:
Confirmed, `AWToggleTriplet` calls `AWToggleSwing` which calls `AWToggleTriplet` again (and the cycle repeats until it crashes).

Fixed in 0cb9e888.

nofishonfriday:
> (It seems odd to me that these actions toggle more than what is announced in their names.)

As odd as it is that Reaper allows both enabled at the same time, but as it does, agreed, it should follow Reaper behaviour (or these actions renamed appropriately) imo.

Breeder:
Told you to not touch it
https://forum.cockos.com/showpost.php?p=2270475&postcount=2791

Revert it to my previous version and I suspect everything will be all right again...

Breeder:
I have my personal SWS build for months which was generated after I pushed this and zero trouble...

Breeder:
> Confirmed, `AWToggleTriplet` calls `AWToggleSwing` which calls `AWToggleTriplet` again (and it recurses infinitely until it crashes).
> 
> Fixed in [0cb9e88](https://github.com/reaper-oss/sws/commit/0cb9e8882be1b55f5e4248e024f7ebd016aef6d3).
> 
> (It seems odd to me that these actions toggle more than what is announced in their names.)

Told you to not touch it
https://forum.cockos.com/showpost.php?p=2270475&postcount=2791

Revert it to my previous version and I suspect everything will be all right again...

Breeder:
If you ever have a question about notation rules, this book is considered a golden standard:
https://www.amazon.com/Music-Notation-Crescendo-Book-Gardner/dp/0800854535

cfillion:
> Told you to not touch it

There has not been a new release yet since December... Neither your original commit nor the modified one(s) has made it to any user.

The recursion bug happens with your original commit. Except for minor refactoring and fixing that crash, I kept your entire code's logic intact. I already reverted all my other changes long ago.

Although I still think `DeferRefreshToolbar` is not ideal. As I wrote above, the toolbars are automatically be refreshed every 300 ms with the default settings. Adding `DeferRefreshToolbar` will cause them to always be refreshed when opening the MIDI editor even when the user disabled the option *Enable toolbars auto refresh*.

> But if it was meant to be like that, it wouldn't be only in arrange. 

It's not only in arrange. It's also available in the MIDI Editor: https://i.imgur.com/ZtBnEvV.png.

Breeder:
I'm talking about the comments in the forum cause that was something I tested and appeared to have been removed, at least judging from the comments. The thing is I have zero times to readily read SWS code - but this exact "bug" appeared when there was no dual defer_toolbar during testing or some action wasn't calling other action and all of these things were critized so I presumed they were removed in official realease, hence my "told you so" comment - this grid thing has been my number 1 enemy in REAPER and I really gave my best to program it so I don't have to return and refix and it was pretty nasty so I tested the hell out of it - I wonder what is the bug this person is experiencing here then. Maybe it's toolbar auto refresh in SWS/S&M part of the code turned off?

nofishonfriday:
>  I wonder what is the bug this person is experiencing here then. 

Quite sure it's because of the changes in Reaper 6.0+ how toggle states are reported.
What that user is reporting works fine here in R5 but not R6.

R5
trip_dot_5

R6
trip_dot_6

PR #1271, opened 2019-12-31
Clean up DockWnd dependency on command Ids

dtormoen:
I'm working on implementing a new module and found that the dependency between action function pointers and the `SWS_DockWnd` constructor made it difficult to avoid a lot of awkward global variables.

As far as I can tell it seems that `m_iCmdID` isn't really used in any useful way so this change seems safe to me. Please let me know if there is something I'm missing!

### DIFF
--- a/Breeder/BR_ContextualToolbars.cpp
+++ b/Breeder/BR_ContextualToolbars.cpp
@@ -2032,7 +2032,7 @@ int BR_ContextualToolbarsView::OnItemSort (SWS_ListItem* item1, SWS_ListItem* it
 * Contextual toolbars window                                                  *
 ******************************************************************************/
 BR_ContextualToolbarsWnd::BR_ContextualToolbarsWnd () :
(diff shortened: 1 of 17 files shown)


cfillion:
Thanks! `m_iCmdID` has not been used in since this commit from 2015: d8be57303568ba7c6ab517b155d344da3ba33be4.

Don't hesitate to squash commits like b4c2db165d88ba02ecdfc58d03eaef3355ab37c5 to keep the history clean. Amending commits and rewriting branch history is fine until the PR is merged.

dtormoen:
You're welcome! Cleaned up the commit history. I'm not sure how you have the repository setup, but I generally like to have Github handle the squashing while merging so I can maintain a branch history until it's merged. I probably should've just done an amend in this case though.

https://help.github.com/en/github/administering-a-repository/configuring-commit-squashing-for-pull-requests

PR #1284, opened 2020-01-09
Improve SWS_GetSelectedTracks, SWS_GetSelectedMediaItems (pre-resize container)

nofishonfriday:
Implemented the suggestion from here: https://github.com/reaper-oss/sws/pull/1230#issuecomment-562285212

### DIFF
--- a/sws_extension.cpp
+++ b/sws_extension.cpp
@@ -717,6 +717,7 @@ error:
 		IMPAPI(CountProjectMarkers);
 		IMPAPI(CountSelectedMediaItems); // O(MN): should be banned from the extension, ideally -- don't use it in loops, at least
 		IMPAPI(CountSelectedTracks); // exclude master + O(N): should be banned from the extension, ideally -- don't use it in loops, at least
+		IMPAPI(CountSelectedTracks2);
 		IMPAPI(CountTakeEnvelopes) // v5pre12+ -- O(N): don't use it in loops
 		IMPAPI(CountTakes);
 		IMPAPI(CountTCPFXParms);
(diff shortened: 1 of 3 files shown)


cfillion:
Nice! The new implementation behaves slightly differently as it no longer clears the array if there are no items/tracks selected. However these functions are never called more than once with the same array in the whole codebase (in fact they're always given a freshly constructed, empty one), so this should not be an issue.

```cpp
WDL_TypedBuf<MediaTrack*> selTracks;
SWS_GetSelectedTracks(&selTracks, false);
assert(selTracks.GetSize() > 0);
Main_OnCommand(40297, 0); // Track: Unselect all tracks
SWS_GetSelectedTracks(&selTracks, false); // doesn't happen in the whole codebase
assert(selTracks.GetSize() == 0); // would have been true before
```

cfillion:
Mhh `GetSelectedTrack2` is actually slower than `CSurf_TrackFromID`+`GetSetMediaTrackInfo`.

With 2,000 tracks and calculating the average runtime of 2,000 calls to SWS_GetSelectedTracks:

- All tracks unselected: from 0.119ms down to 0.019ms 🎉 
- All tracks selected: it increased from 0.16ms to 15ms! 🐢 
- With the old loop code but a single Resize: no noticeable difference (0.15ms)

<details>
<summary>Test code</summary>

```cpp

void SWS_GetSelectedTracks(WDL_TypedBuf<MediaTrack*>* buf, bool bMaster)
{	
	int selTracksCount = CountSelectedTracks2(nullptr, bMaster);
	if (selTracksCount == 0)
		return;

	buf->Resize(selTracksCount);
	for (int i = 0; i < selTracksCount; i++)
	{
	  if (*(int*)GetSetMediaTrackInfo(tr, "I_SELECTED", NULL))
	  {
	    int pos = buf->GetSize();
	    buf->Get()[pos] = tr;
	  }
	}
}
```
</details>

nofishonfriday:
@cfillion Thanks for benchmarking.

> Sorry, looks like my suggestion was a bad idea.
The results are quite surprising though, imo.
Closing.

Issue #1285, opened 2020-01-09
Resourses : bugs 


mehmethand:
Bugs:
1)" Double click to load selected fx chain"  floats the plugins which were floating while saving the fxchain. This is expected behaviour. But "draging selected fxchain into arrange window" make all plugins float  and fx chain window open. Is it possible to make dragging behave like double-click?
Request

2) please take a look at attached picture . I typed "fab eq" on filter, and I'm expecting to see the results as "fx chains which contains both fab and eq in the name or comments". But results show anything that contains fab or eq in the text. Please make this behaivour like action filter or media explorer filtersç
![resources](https://user-images.githubusercontent.com/16308632/72112906-03a88f00-3350-11ea-9043-969204daf3e6.jpg)

Request:
 Focus on "filter" search text area when opened  like fx browser.

Regards...

Reaper 6.02 , sws 2.11.0 win  x64

Issue #1257, opened 2019-12-15
auto color: exact name match


StephanRoemer:
It would be great to have the ability to exactly match the track name in the auto color dialog.

Example: 
I have a main folder that is called "PERCUSSION", the sub folders are "Tuned Percussion", "Untuned Percussion" and "Epic Percussion". So there's a problem with the name matching, because PERCUSSION is found 4 times and overwrites the color setting. 

Maybe a feature to set the name in "", like "PERCUSSION" to exactly match the name, would help here? Case matching as well as string length.

Issue #1119, opened 2019-03-04
Several improvements for the Resource window


StephanRoemer:
- The search field does not work as expected. When you enter two key words, the results don't get narrowed down. Sometimes even the opposite is the case and you see more results than with only one keyword.
https://i.imgur.com/x58Y7J7.gif

In the GIF above, I would have expected that entering "sym" and "soft" would only return the
"SYM Strings Sus Extra Soft" patch.

- the DEL key has no effect in the search field
- it would be great if ENTER would jump immediately to the list, so that we can select the track template with the cursor keys.

nofishonfriday:
> The search field does not work as expected. When you enter two key words, the results don't get narrowed down. Sometimes even the opposite is the case and you see more results than with only one keyword.

That's because a space in the filter means OR:
results = term1 OR term2...

I've added supporting AND keyword (like in Reaper's action list), so if you type in _sym AND soft_ it should now only show result "SYM Strings Sus Extra Soft".

If you want to test:
https://www.dropbox.com/s/fqd5ei9gm1ep639/reaper_sws64.dll?dl=1

StephanRoemer:
Thanks nofish, this works as expected! Thanks so much!

Just a little detail: would it be possible to make AND work as default mode (without the need to type it)? That's how the action list works as well. AND is default and if we need OR, we can type it.

nofishonfriday:
Thanks for testing and reporting back.

About making AND default, hm.. I fear someone will complain about changing default behaviour then. :P
I'll think about it...

StephanRoemer:
Hmm maybe you could make it an option (checkbox) to select the default behavior?

StephanRoemer:
On the other hand, I can't think of any default behavior where adding several words would result in an OR search. Even the Media Explorer has AND as default. Just as general thought. Maybe it would be more consistent to apply the same behavior to the resources window.

nofishonfriday:
> On the other hand, I can't think of any default behavior where adding several words would result in an OR search. Even the Media Explorer has AND as default. Just as general thought. Maybe it would be more consistent to apply the same behavior to the resources window.

On further thought, agreed. :)
cfillion has a nicely working filter (which I think defaults to AND as should be) implemented in ReaPack, maybe we can reuse that at some point.

StephanRoemer:
That would be awesome! :)

X-Raym:
@nofishonfriday Hi ! How this has been implemented ? It is only OR filter for now.

[...]

PR #1291, opened 2020-01-17
Add regex support in Auto Color for track, marker, region name filters

daleonov:
Exact matching didn't really work for track names for me. So I added regex support to define a pattern, and keywords like "gtr", "bass", "synth" won't conflict with "lead", "rhythm", "double" etc and each combo would be able to have its own color.
sws regex

### DIFF
--- a/Color/Autocolor.cpp
+++ b/Color/Autocolor.cpp
@@ -34,6 +34,7 @@
 #include "../SnM/SnM_Util.h"
 #include "../reaper/localize.h"
 #include "WDL/projectcontext.h"
+#include <regex>
 
 #define PRI_UP_MSG         0x10000
 #define PRI_DOWN_MSG       0x10001
@@ -86,6 +87,8 @@ static bool g_bAIEnabled = false;
 static bool g_bALEnabled = false;
 static WDL_String g_ACIni;
 static int s_ignore_update;
+static const char g_cRegexPrefix = '$'; // Filter should start with that prefix to be treated as regex
+static const auto g_bmRegexOptions = std::regex_constants::extended | std::regex_constants::icase | std::regex_constants::collate;
 
 
 // Register to marker/region updates
@@ -812,8 +815,20 @@ void ApplyColorRuleToTrack(SWS_RuleItem* rule, bool bDoColors, bool bDoIcons, bo
 					else // Check for name match
 					{
 						char* cName = (char*)GetSetMediaTrackInfo(tr, "P_NAME", NULL);
-						if (cName && stristr(cName, rule->m_str_filter.Get()))
-							bMatch = true;
+						if (cName){
+							if (stristr(cName, rule->m_str_filter.Get())){
+								// Exact match
+								bMatch = true;
+							}
+							else if (strlen(rule->m_str_filter.Get()) > 1 && (rule->m_str_filter.Get())[0] == g_cRegexPrefix){
+								// Regex pattern match
+								const char* sUserPattern = (char*) rule->m_str_filter.Get() + 1; // Chopping the prefix
+								std::regex reUserPattern(sUserPattern, g_bmRegexOptions);
+
+								if (std::regex_match(cName, reUserPattern))
+									bMatch = true;
+							}
+						}
 					}
 				}
 				else if (strcmp(rule->m_str_filter.Get(), cFilterTypes[AC_MASTER]) == 0)
@@ -1064,6 +1079,7 @@ void ApplyColorRuleToMarkerRegion(SWS_RuleItem* _rule, int _flags)
 	int x=0, num, color;
 	bool isRgn;
 	const char* name;
+	const bool isRegex = (strlen(_rule->m_str_filter.Get()) > 1) && ((_rule->m_str_filter.Get())[0] == g_cRegexPrefix);
 
 	PreventUIRefresh(1);
 	if(_rule->m_type & _flags)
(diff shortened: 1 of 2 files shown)


daleonov:
I can also look into #1257 since it's about similar feature.

nofishonfriday:
I was unaware of this either until [shortly ago](https://github.com/reaper-oss/sws/pull/1230#issuecomment-562285212) but since SWS targets macOS 10.5. the C++11 regex features are not available for the Mac version, so we can't use them unfortunately (afaik).
https://travis-ci.com/reaper-oss/sws/jobs/276788893#L546

cfillion:
In addition to the above, the `std::regex` constructor throws an `std::invalid_regex` exception when the input is invalid. Since it comes from users, the possibility of an invalid regex should be handled without crashing.

Currently it bricks REAPER by making it crash on startup until the invalid regex is manually removed from the .ini file.

I find the `$` prefix somewhat confusing too, as it is usually the end of string/line anchor. Perhaps surrounding the regexes with slashes would make it clearer? It is a widely used delimiter for regexes.

Additional concerns I'm not sure about (needs testing):
- Cost of compiling regexes over and over with lots of tracks
- Possible hangs with input accidentally resulting in overly complex regexes?
- What are the advantages of using the POSIX extended syntax vs the default ECMAScript?

As for implementing #1257 (and the related #1119), that can be done by reusing ReaPack's filter implementation (but with its use of the C++17 stdlib removed) as discussed [here](https://github.com/reaper-oss/sws/issues/1119#issuecomment-493083501).

daleonov:
> the C++11 regex features are not available for the Mac version, so we can't use them unfortunately (afaik).

Can't use them at all or still can use as a platform-specific feature? The code is easy to isolate so I can do something like
```
#ifndef GRANDPA_TECH
do_regex();
#endif 
```
...if such thing is acceptable here, of course.
Other way is using third-party library, but I don't think it's welcome here unless extremely necessary?

> the std::regex constructor throws an std::regex_error exception when the input is invalid
Fair enough. No elegant way to address it other than catching it? Good news is `std::regex_error` is easy to isolate.

> Perhaps surrounding the regexes with slashes would make it clearer?
Good idea!

> Cost of compiling regexes over and over with lots of tracks (vs storing a compiled `std::regex` and reusing it)
Fair enough. It's a lower priority concern than the rest of those. But if I go with regex exception handling (as I discussed above) then I'll have one more reason to move it outside the loop.

> Possible hangs with user input accidentally resulting in overly complex regexes?
Possible... But it's their DAW, right? I think it's up to them to decide how complex is too complex :)

> What are the advantages of using the POSIX extended syntax vs the default ECMAScript?
There was something I didn't like, so I always used the extended one. Don't remember exactly, I'll check later. As long as it can do what's on screenshot in first post, I'm happy.

> Wouldn't the `collate` option introduce behavior differences from an install to the next depending on the current system locale?
And I think it should. What I had in mind is stuff like `\w` including user's alphabet. I haven't tested it yet but I assumed it's what this flag does.

Also, I've hardcoded it to be case insensitive, not quite happy with it. I don't recall anything in actual regex syntax to specify case sensitivity, and I don't see any elegant way to specify it in the sws enviroment. Other than adding another meta language thing to filter string, but that's clunky too.

----

Okay, so let's first decide if it's possible at all to add this feature (since there are some macOS compatibility issues) and if it is, I'll proceed with addressing the points above. Hopefully I won't have to fork it just for that little feature :)

cfillion:
> Can't use them at all or still can use as a platform-specific feature?

It wouldn't be the first platform-specific feature in SWS (there are some Windows-only ones). It would be a shame to have this everywhere except on macOS though...

> Other way is using third-party library, but I don't think it's welcome here unless extremely necessary?

Pretty much. Regexes in auto color would be a pretty niche feature so I don't think it's enough to warrant bundling an entire regex library. From what I could find (ignoring compatibility concerns regarding the regex flavor), Apple's Foundation framework added regexes in v10.7 and libpcre is bundled with the OS since 10.6. 10.5 doesn't get any.

> No elegant way to address it other than catching it?

That's correct.

> Possible... But it's their DAW, right?

If they can accidentally brick their DAW (eg. freeze at startup), it might become an issue. (I haven't looked into it so it might be very hard/impossible to so. In which case this is a non-issue.)

> And I think it should. What I had in mind is stuff like \w including user's alphabet.

That would cause issues when using an existing config one a computer that happens to have a different locale than the one it was tested on (eg. cool Auto Color regex tips shared on the forum mysteriously doesn't work, portable install on a thumb drive appears broken at the studio's PC...).

> I don't recall anything in actual regex syntax to specify case sensitivity, and I don't see any elegant way to specify it in the sws enviroment.

How about `/regex string/i` for case-insensitive? Since it's a common syntax used in Perl, JavaScript, Ruby, PHP...

---

Alternatively, instead of regular expressions, perhaps REAPER's Action List syntax would be helpful in Auto Color? It has AND/OR/NOT, grouping with parenthesis, and full-word matching.

It's not very useful alternative with patterns such as `V[0-9]` (becoming the very ugly `V0 OR V1 OR V2 OR...`), but things like `(GTR|guitar).*lead` can be written as `( GTR OR guitar ) lead` (except it allows `lead` to appear anywhere including before `GTR`).

daleonov:
Using simpler syntax would be great! I agree, regex is a bit too geeky even for reaper users.
But parsing it without regex and not being able to use regex under the hood would be one hell of a chore... Is something similar already implemented somewhere in sws?

cfillion:
In that case this PR, along with #1119 (where it was originally suggested), #1257 and #1285 could all be solved by adding ReaPack's [implementation](https://github.com/cfillion/reapack/blob/master/src/filter.cpp) of it into SWS (with minor tweaks to remove uses of `std::unique_ptr` and `boost::algorithm::to_lower`) .

odedd:
Any chance of getting this thing to work? This could be a really powerful addition

sychron:
Unfortunately, My c++ FU is not good enaugh to resolve the conflicts. 

But I would really like to have this feature, as it's very handy if you have track names that have the description in the front and a labelling in the back, for example: "FX: <rolename> A", "FX: <rolename> B" etc. With Regex, one could assign track icons like FX:A, FX:B, ignoring the role name.
So ... please count this as an upvote for this request :)

PR #1294, opened 2020-01-18
Prevent creating SWS_Global notes.txt if global notes feature isn't used

nofishonfriday:
report:
https://forum.cockos.com/showpost.php?p=2233645&postcount=2755

### DIFF
--- a/SnM/SnM_Notes.cpp
+++ b/SnM/SnM_Notes.cpp
@@ -95,7 +95,7 @@ SWSProjConfig<WDL_FastString> g_prjNotes; // extra project notes
 // global notes #647, saved in <REAPER Resource Path>/SWS_GlobalNotes.txt 
 // (no SWSProjConfig, one instance across all projects, i.e. global)
 WDL_FastString g_globalNotes; 
-bool g_IsGlobalNotesDirty = false;
+bool g_globalNotesDirty = false;
 
 int g_notesType = -1;
 int g_prevNotesType = -1;
@@ -620,7 +620,7 @@ void NotesWnd::DrawControls(LICE_IBitmap* _bm, const RECT* _r, int* _tooltipHeig
 					}
 					break;
 				case SNM_NOTES_GLOBAL:
-					if (g_IsGlobalNotesDirty)
+					if (g_globalNotesDirty)
 						lstrcpyn(str, "[modified]", sizeof(str));
 					else
 						lstrcpyn(str, "", sizeof(str));
@@ -839,7 +839,7 @@ void NotesWnd::SaveCurrentGlobalNotes(bool _wantUndo)
 		MarkProjectDirty(NULL);
 	*/
 
-	g_IsGlobalNotesDirty = true;
+	g_globalNotesDirty = true;
 	RefreshGUI(); // to display [modified] in notes window immediately
 }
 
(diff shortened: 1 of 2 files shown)


SnM/SnM_Notes.cpp
```diff
@@ -1472,17 +1472,29 @@ void WriteGlobalNotesToFile()
...
+		{
+			MessageBox(GetMainHwnd(), __LOCALIZE("Saving global notes failed.", "sws_mbox"), __LOCALIZE("SWS - Error", "sws_mbox"), MB_OK);
+		}
 	}
-	else // writing failed
+	else // remove if empty
 	{
-		MessageBox(GetMainHwnd(), __LOCALIZE("Saving global notes failed.", "sws_mbox"), __LOCALIZE("SWS - Error", "sws_mbox"), MB_OK);
-	}
+		std::remove(filePath.Get());
```
cfillion: Mhh I'm pretty sure this fails on Windows if the file path contains characters outside of the current system ANSI codepage.

If the file must be removed when empty, using the wide version of `remove` should do the trick:
```suggestion
#ifdef _WIN32
		_wremove(win32::widen(filePath.Get()).c_str());
#else
		remove(filePath.Get());
#endif
```

nofishonfriday: Thanks for reviewing.
> If the file must be removed instead of saved empty
I'm not very confident with it myself (think empty text file is not that much harm), but it's in reaction to [edgemeal's comment](https://forum.cockos.com/showpost.php?p=2233645&postcount=2755) the file appearing despite not using this feature. 

What I particularly don't like, since global notes are saved on project save, it now tries to delete this file every time, also when it's not there (user hasn't used the feature). I could check before if the file actually exists but I guess that's not more efficient than calling `remove()` directly, as this also checks the existence first anyway?
But then again there should be a way to handle when user purposely wants to save empty notes (to wipe previous ones) so I think I can't omit the `g_globalNotes.GetLength() == 0` condition completely.
Do you by chance have a better idea how to handle it?

nofishonfriday: Found a better way (I think).
Checking for `g_IsGlobalNotesDirty` before saving prevents creating the SWS_Global Notes.txt file if user isn't using this feature (and no need for std::remove anymore).

cfillion: Good idea! (I was going to suggest using a smart pointer for `g_globalNotes` defaulting to null if the user didn't use global notes.)

Could the notes be saved when closing the notes window (eg. from `NotesWnd::OnDestroy`)? Is there a reason saving any project is required (or using the menu)?

Also I found a bug: because `WriteGlobalNotesTofile` uses `WDL_FastString::get_filepart` instead of `Get`, everything before (and including) the last slash is discarded and garbage from outside of the buffer is appended (since it's still reading `GetLength` worth of characters).

(nitpick: the extra indentation in `WriteGlobalNotesToFile` can be avoided by returning early: `if (!g_globalNotesDirty) return;`)

nofishonfriday: Thanks!
>  I found a bug:

Should be fixed, just force-pushed.

> nitpick:
Also done.

> Could the notes be saved when closing the notes window (eg. from NotesWnd::OnDestroy)? Is there a reason saving any project is required (or using the menu)?

Why does Reaper provide a 'Save' menu item instead of just saving automatically when quitting? :P
More seriously, I actually regarded it as a feature providing a means to the user saving the global notes programmatically and not as a requirement (i.e. hassle).
Case :): user  is in progress writing his/her 10.000+ chars global notes, Reaper crashes while doing something else intermittedly, notes would be lost if not saved before (`NotesWnd::OnDestroy` wouldn't be called in this case).
And I find the most expected/user friendly way is that 'stuff' gets saved with a project save (I regard the saving from notes window window just as 'sugar on top').

Would that be a valid argument you think? :)

cfillion: > Would that be a valid argument you think? :)

Yes, thanks for explaining!

> Why can't I use `std::strlen` on Linux (and `strlen` works)?

`std::strlen` is defined in the `cstring` standard C++ header. SWS explicitly includes the C variant (`string.h`) which may or may not put functions in `std`. `cstring` is probably getting included by another header on macOS and Windows.

cfillion: Just read the new patch. The buffer over-read is fixed, but shouldn't it save the whole thing that the user typed rather than only what's after the last slash/backslash?

```cpp
const int wrtPos = outfile.Write(g_globalNotes.Get(), g_globalNotes.GetLength());
```

nofishonfriday: Oh, somehow I got totally confused with the `get_filepart` function. I somehow thought I'd need it for saving the file, just looked at the definition now what it actually does. :1st_place_medal: 

So just:
```
const int wrtPos = outfile.Write(g_globalNotes.Get(), g_globalNotes.GetLength());
if (wrtPos == g_globalNotes.GetLength()) // writing OK
{ ...
```

right?
Thanks.

cfillion: Yes (either way is fine due to `WDL_FastString::GetLength`'s constant-time complexity. `globalNotesLength` should probably be an `int` to match the return type of `GetLength`).

PR #1304, opened 2020-02-01
Theme helper: add support for REAPER v6

nofishonfriday:
report (and confirmed here): https://forum.cockos.com/showthread.php?t=230558

alternative way to get TCP width (than reading ftom .ini, as it's not in the API obviously) would be:
```
#include "Breeder/BR_Util.h" // GetTcpWnd()
bool isContainer;
HWND tcpWnd = GetTcpWnd(isContainer);
RECT r;	GetClientRect(tcpWnd, &r);
int tcpwWdth = r.right - r.left;
```

but from my (and [Edgemeal's](https://forum.cockos.com/showpost.php?p=2238757&postcount=5)) test both give same result, so not sure which one is better.

It loops through the tracks two times, could have been optimized probably, but I wanted to keep the previous output structure (all TCPs first, then MCPs) and this way it was easier to do. :)

edit:
I left the old code in currently. `I_TCPH` was added in [v5.981+dev0803](https://forum.cockos.com/showthread.php?p=2166101), but I couldn't find when `I_MCPH/I_MCPW` were added. If before/at v5.982 (next SWS required min. REAPER version) the legacy code could theoretically be removed.

### DIFF
--- a/SnM/SnM_Window.cpp
+++ b/SnM/SnM_Window.cpp
@@ -32,6 +32,7 @@
 #include "SnM_Track.h"
 #include "SnM_Util.h"
 #include "SnM_Window.h"
+#include "Breeder/BR_Util.h" // GetTcpWnd()
 #include "../reaper/localize.h"
 
 
@@ -371,8 +372,8 @@ void ShowThemeHelper(WDL_FastString* _report, HWND _hwnd, bool _mcp, bool _sel)
 							1024,
 							__LOCALIZE_VERFMT("%s Track #%d '%s': W=%d, H=%d\n","theme_helper"),
 							_mcp ? __LOCALIZE("MCP","theme_helper") : __LOCALIZE("TCP","theme_helper"),
-							trIdx==-1 ? 0 : trIdx,
-							trIdx==-1 ? __LOCALIZE("[MASTER]","theme_helper") : (trName?trName:""),
+							trIdx,
+							trIdx==0 ? __LOCALIZE("[MASTER]","theme_helper") : (trName?trName:""),
 							(int)(r.right-r.left),
 							(int)(r.bottom-r.top));
 					}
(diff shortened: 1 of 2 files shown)


cfillion:
`I_MCP[XYWH]` were added in REAPER v6.0.

SnM/SnM_Window.cpp
```diff
@@ -382,20 +382,90 @@ void ShowThemeHelper(WDL_FastString* _report, HWND _hwnd, bool _mcp, bool _sel)
...
+	int tcpWidth = 0;
+	if (!_mcp)
+	{
+		tcpWidth = GetPrivateProfileInt("REAPER", "leftpanewid", -666, get_ini_file());
+		// leftpanewid= entry gets created first time user drags TCP/arrange divider
+		if (tcpWidth == -666)
+		{
+			MessageBox(
+				g_hwndParent, __LOCALIZE("'leftpanewid=' entry to determine TCP width doesn't exist in reaper.ini.\nPlease drag vertical TCP/arrange devider once to create it then run theme helper actions again.", "sws_mbox"), 
+				__LOCALIZE("Theme helper - Error", "sws_mbox"), MB_OK);
```
cfillion: I think the `GetClientRect` approach is better because it works out of the box.

SnM/SnM_Window.cpp
```diff
@@ -382,20 +382,90 @@ void ShowThemeHelper(WDL_FastString* _report, HWND _hwnd, bool _mcp, bool _sel)
...
+			return;
+		}
+	}
+		
+	for (int trIdx = 0; trIdx <= GetNumTracks(); trIdx++)
+	{
+		{
```
cfillion: Why the extra brace/indentation level? The `for` loop is already a new scope for the variables inside.

Since both the `_mcp` and `!_mcp` branches are mostly the same, this can be shortened into (and as a bonus, this keep the existing translations of the Rv5 format string):

```cpp
static int GetTcpWidth()
{
	bool isContainer;
	HWND tcpWnd = GetTcpWnd(isContainer);
	RECT r;	GetClientRect(tcpWnd, &r);
	return r.right - r.left;
}

void ShowThemeHelperREAPERv6(WDL_FastString* _report, bool _mcp, bool _sel)
{
	const int tcpWidth = _mcp ? 0 : GetTcpWidth();

	for (int trIdx = 0; trIdx <= GetNumTracks(); ++trIdx)
	{
		MediaTrack* tr = CSurf_TrackFromID(trIdx, false);

		if (_sel && !GetMediaTrackInfo_Value(tr, "I_SELECTED"))
			continue;

		const char* trName = static_cast<const char *>(GetSetMediaTrackInfo(tr, "P_NAME", nullptr));
		if (!trName) trName = "";

		const int width = _mcp ? static_cast<int>(GetMediaTrackInfo_Value(tr, "I_MCPW")) : tcpWidth;
		const int height = static_cast<int>(GetMediaTrackInfo_Value(tr, _mcp ? "I_MCPH" : "I_TCPH"));

		if (!height) // master not visible or track hidden
			continue;

		_report->AppendFormatted(1024,
			__LOCALIZE_VERFMT("%s Track #%d '%s' : W=%d, H=%d\n", "theme_helper"),
			_mcp ? __LOCALIZE("MCP", "theme_helper") : __LOCALIZE("TCP", "theme_helper"),
			trIdx, trIdx == 0 ? __LOCALIZE("[MASTER]", "theme_helper") : trName,
			width, height
		);
	}
}
```

cfillion:
This is not new, but the Rv5 implementation has a bug:

https://github.com/reaper-oss/sws/blob/504455a3a3d55314ac6586126ac129696068d262/SnM/SnM_Window.cpp#L374-L375

`trIdx` can never be `-1`:

https://github.com/reaper-oss/sws/blob/504455a3a3d55314ac6586126ac129696068d262/SnM/SnM_Window.cpp#L366

Therefore it shows an empty track name instead of "MASTER".

PR #1318, opened 2020-03-09
Snapshots: store/recall track playback offset (REAPER v6.0+)

### LINKED ISSUE #1313: FR: Apply Timing Offset from IO Window
Snapshots should also save the Timing Offset specified in the IO Window.

nofishonfriday:
closes #1313

Contains code which (hopefully) should make it compatible with older Reaper version (though the feature wouldn't be available there obviously, as it was introduced in v6.0. The related tickbox is always disabled/unticked in this case.)


Snapshots/Snapshots.cpp
```diff
@@ -831,7 +840,10 @@ void SWS_SnapshotsWnd::GetOptions()
 	if (IsDlgButtonChecked(m_hwnd, IDC_MIX) == BST_CHECKED)
 	{
 		m_iSelType = 0;
-		g_iMask = MIX_MASK;
+		if (atof(GetAppVersion()) < 6)
+			g_iMask = MIX_MASK_NO_OFFS;
+		else
+			g_iMask = MIX_MASK;
```
cfillion: How about something like this here and in `SetSnapType`? (In the case of new future features having to be disabled according to the running REAPER version, having the check(s) in one place should be slightly easier to maintain.)

```cpp
g_iMask = MIX_MASK & g_compatMask;
```

Where `static int g_compatMask = ALL_MASK`, and in `SnapshotsInit`:
```cpp
// disable features unavailable in the running version of REAPER
if (atof(GetAppVersion()) < 6)
  g_compatMask &= ~PLAY_OFFSET_MASK;
```

sws_extension.rc
```diff
@@ -140,7 +140,8 @@ BEGIN
     CONTROL         "Sends",IDC_SENDS,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,171,91,35,10
     CONTROL         "Vis",IDC_VISIBILITY,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,140,103,27,10
     CONTROL         "Selection",IDC_SELECTION,"Button",BS_AUTOCHECKBOX | WS_TABSTOP,171,103,45,10
-    CONTROL         "Phase", IDC_PHASE, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 140, 115, 45, 10
+    CONTROL         "Pol", IDC_PHASE, "Button", BS_AUTOCHECKBOX | WS_TABSTOP, 140, 115, 32, 10
```
cfillion: Changing Phase to Pol(arity) although it's still labeled as "Phase" in REAPER and other areas of SWS and Snapshots?

nofishonfriday: Uhm that was purely because Phase didn't quite fit anymore. :)
!(https://i.imgur.com/1YDudqp.jpg)

Any suggestion?

cfillion: There's unused room to the right, the second column can be moved a little bit. The panel could be made a little bigger to accommodate the new contents if needed.



(There's room for writing "Visibility" in full by moving it in place of the Sends checkbox)

nofishonfriday:
While at Snapshots window, there's a small issue it not opening at full size when opening for the very first time.
How it looks here with fresh Reaper/SWS install opening for the first time:


But I haven't found yet where to tweak the initial size.
Once resized it ok (because it does store size/pos.)

cfillion:
The default size is set in the first line of the dialog definition in the resource file:

https://github.com/reaper-oss/sws/blob/08d3c6fce05191347092a55df6dd5d4764f1de9c/sws_extension.rc#L117

IDC_LIST's height (the last parameter in its line) must be adjusted by the same amount as the window height.

Snapshots/Snapshots.cpp
nofishonfriday: I like the elegantness of this. Thanks.

nofishonfriday:
I think now I could also do something like for the compatibility check:

in `SWS_SnapshotsWnd::Update()`, instead of
https://github.com/reaper-oss/sws/pull/1318/files#diff-3aee35788be6fa8fd616961d2252ce24R425-R433

```
	for (int i = 0; i < MASK_CTRLS; i++)
	{
		if (i == 9 && ALL_MASK - g_compatMask == PLAY_OFFSET_MASK)
			continue;
		CheckDlgButton(m_hwnd, cSSCtrls[i], g_iMask & cSSMasks[i] ? BST_CHECKED : BST_UNCHECKED);
		EnableWindow(GetDlgItem(m_hwnd,  cSSCtrls[i]), m_iSelType == 2);
	}
```

Does this make sense?

sws_extension.rc
nofishonfriday: > There's enough space for writing "Visibility" in full by moving it in place of the Sends checkbox too

I've done that but on rethought, I hope noone will complain about wasted muscle memory if the tickboxes are swopped?

cfillion:
> Is this one correct?

Yes: `!(g_compatMask & PLAY_OFFSET_MASK)` is true when `g_compatMask` does not have any of the bits in `PLAY_OFFSET_MASK` set.

sws_extension.rc
cfillion: Hmm yeah, and it would not match the checkboxes in the "SWS Snapshot Paste" dialog anymore. I think it's clearer and looks better than the abbreviated form though.

The paste dialog should be updated to match the changes in the main snapshots window. The new Phase feature is missing there by the way.

nofishonfriday: > The paste dialog should be updated to match the changes in the main snapshots window. The new Phase feature is missing there by the way.

Thanks.

Snapshots/Snapshots.cpp
```diff
@@ -421,6 +422,12 @@ void SWS_SnapshotsWnd::Update()
 		CheckDlgButton(m_hwnd, cSSCtrls[i], g_iMask & cSSMasks[i] ? BST_CHECKED : BST_UNCHECKED);
 		EnableWindow(GetDlgItem(m_hwnd,  cSSCtrls[i]), m_iSelType == 2);
```
cfillion: The check below could be done here so that it gets applied to any future version-specific features without code duplication:
```suggestion
		CheckDlgButton(m_hwnd, cSSCtrls[i], g_iMask & cSSMasks[i] & g_compatMask ? BST_CHECKED : BST_UNCHECKED);
		EnableWindow(GetDlgItem(m_hwnd,  cSSCtrls[i]), m_iSelType == 2 && cSSMasks[i] & g_compatMask);
```

Issue #1335, opened 2020-04-27
BUG: Loudness doesn't work for items with take volume automation


teeramusic:
None of the loudness features work for items with take volume automation

nofishonfriday:
Which SWS version are you using?
Current beta contains a fix* for this.
https://www.sws-extension.org/download/pre-release/

*added in v2.10.0:
>Fix item Loudness analysis if take contains vol. envelope and item position not 0.0

teeramusic:
I'm using the latest stable (2.10.0 #1 Feb 6, 2019) but I'll try grabbing the beta. Will using the beta make my reaper less stable?

nofishonfriday:
Closing because likely fixed in v2.10.0 and no feedback from OP.

PR #1338, opened 2020-04-30
Loudness: fix momentary calculation

nofishonfriday:
Came across this when looking into https://github.com/reaper-oss/sws/issues/1335.
current line 
`if (i % 2 && momentaryFilled && doMomentary)` (momentaryFilled is bool)
can evaluate differently than Breeder's original line
`if (i % 2 == momentaryFilled && doMomentary)` (momentaryFilled is int)
e.g. when i = 0; momentaryFilled = false/0 ; doMomentary = true -> 1st line == false, 2nd line == true

With the cast to int the "unsafe conversion from bool to int" warning (https://github.com/reaper-oss/sws/pull/1215#issuecomment-545528589) could be avoided but then it could be thought if maybe better that momentaryFilled stays an int rather than casting every time.

@cfillion 
What do you think?

`

### DIFF
--- a/Breeder/BR_Loudness.cpp
+++ b/Breeder/BR_Loudness.cpp
@@ -1153,7 +1153,7 @@ unsigned WINAPI BR_LoudnessObject::AnalyzeData (void* loudnessObject)
 			}
 
 			// Momentary buffer (400 ms) filled
-			if (i % 2 && momentaryFilled && doMomentary)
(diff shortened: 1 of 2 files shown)


cfillion:
Ah, right. I think either `momentaryFilled` should be reverted to an `int` or `i % 2` converted into a boolean (`i % 2 > 0`).

```cpp
if (i % 2 > 0 == momentaryFilled && doMomentary)
```

PR #1339, opened 2020-04-30
Autorender: fix writing metadata

nofishonfriday:
report; https://forum.cockos.com/showthread.php?p=2280226#post2280226 (confirmed here)

@srstclair 
Appreciated If you're avasilable for taking a look. :)
As far as I understand the code (and tested) it seems to be a simple fix, I might miss something though...


shane-axiom:
I haven't looked at this code in almost 10 years, but your fix seems sane to me. Thanks!

PR #1343, opened 2020-05-05
slightly clean up BR_EnvelopeUtil.h/.cpp

nofishonfriday:
slight refactor/clean up of `BR_EnvelopeUtil.h/.cpp` according to schwa's info:
https://forum.cockos.com/showthread.php?p=2198426&postcount=118

- deprecate/remove m_properties.visUnknown (is always 1.0)
- rename m_properties.shapeUnknown1/2 to m_properties.pitchEnvRange/pitchEnvSnap (only used class internal, so should be backwards compatibel)

### DIFF
--- a/Breeder/BR_EnvelopeUtil.cpp
+++ b/Breeder/BR_EnvelopeUtil.cpp
@@ -1793,7 +1793,7 @@ bool BR_Envelope::FillProperties () const
 					lp.parse(token);
 					m_properties.visible    = lp.gettoken_int(1);
 					m_properties.lane       = lp.gettoken_int(2);
-					// 3rd field is a deprecated value, always 1.0, see p=2198426&postcount=118
+					// 3rd field is a deprecated value, always 1.0, see p=2198426
 				}
 				else if (!strncmp(token, "LANEHEIGHT ", sizeof("LANEHEIGHT ")-1))
 				{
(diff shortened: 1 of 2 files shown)


Breeder/BR_EnvelopeUtil.cpp
```diff
@@ -1908,10 +1910,10 @@ WDL_FastString BR_Envelope::GetProperties ()
 		properties.Append(m_properties.paramType.Get());
 		properties.Append("\n");
 		properties.AppendFormatted(256, "ACT %d %d\n", m_properties.active, m_properties.AIoptions);
-		properties.AppendFormatted(256, "VIS %d %d %d\n", m_properties.visible, m_properties.lane, m_properties.visUnknown);
+		properties.AppendFormatted(256, "VIS %d %d %d\n", m_properties.visible, m_properties.lane, 1.0);
```
cfillion, line 1913:
```suggestion
		properties.AppendFormatted(256, "VIS %d %d %d\n", m_properties.visible, m_properties.lane, 1);
```

`1.0` must be an integer to match the `%d` format. Otherwise garbage is inserted into `properties`. I'll force-push a commit to fix this and amend the 2.12.1 branch with it.

```
../Breeder/BR_EnvelopeUtil.cpp:1933:94: warning: format specifies type 'int' but the argument has type 'double' [-Wformat]
                properties.AppendFormatted(256, "VIS %d %d %d\n", m_properties.visible, m_properties.lane, 1.0);
                                                           ~~                                              ^~~
                                                           %f
```

PR #1374, opened 2020-06-10
Fix/refactor "Rename take source files"/"Rename takes and source files" not working if source is reversed/section

nofishonfriday:
Report: https://forum.cockos.com/showpost.php?p=2300847&postcount=2840
Other changes:
- combine both actions into one function (was almost identical code)
- restore offline state of previously (user-) offlined takes
- display error message if file(s) couldn't be renamed (and abort source renaming)
- t_renameparams struct: use C++ style and init to default


nofishonfriday:
Not happy how it turned out so far, got more and more convoluted.
I'll reopen if I come up with something better.
@cfillion If you're motivated to do it, fine with me also, otherwise I keep trying. :)

Issue #966, opened 2018-03-26
SWS track height actions should obey track height locking


nofishonfriday:
In R5.78 track height locking was added.
SWS actions manipulating track height should leave tracks with locked height untouched too I think.

So far came up these actions in the pre forum which 'broke' in this regard:
- Xenakios/SWS: Set selected tracks heights to A/B
- SWS: Vertical zoom to selected tracks (and the variants of it)

No API currently to Get/Set track height locked state (afaik) but seems it would be possible via chunk parsing at least (In TRACK chunk, (newly introduced ?) 3rd field in the TRACKHEIGHT entry seems to indicate height locked state).

PR #1346, opened 2020-05-08
Fix vertical zoom actions disable track height lock, Add action "SWS/NF: Toggle track height - and vertical zoom actions obey track height lock"

nofishonfriday:
fixes #1221, implements #966 (optionally)
The toggle action was added so [users who prefer the current behaviour](https://forum.cockos.com/showthread.php?p=2091347#post2091347) (Xenakios track height actions and SWS vertical zoom actions ignore track height locking) can have that too.

+revert adding actions "Xenakios/SWS/NF: Set selected tracks heights to A/B, respect height locked tracks" (added in v2.12.0pre)
+refactor "Xenakios/SWS: Set selected tracks heights to A/B" (use COMMAND_T::user)

PR #1414, opened 2020-10-08
Fix vertical zoom actions disable track height lock (fixes #1221), Add action "SWS/NF: Toggle track height - and vertical zoom actions obey track height lock" (implements #966), defaults to off to be consistent with previous SWS versions

### LINKED ISSUE #1221: Vertical zoom actions disable track height lock
reminder for:
https://forum.cockos.com/showthread.php?t=205372

I think issue is in `Zoom.cpp/VertZoomRange()` (maybe other places as well, didn't check)
https://github.com/reaper-oss/sws/blob/221e764c213efb8f0f0def93e07e021c17d153ab/Zoom.cpp#L131
https://github.com/reaper-oss/sws/blob/221e764c213efb8f0f0def93e07e021c17d153ab/Zoom.cpp#L263

as this disables the height lock.
https://forum.cockos.com/showthread.php?p=2202082#post2202082

nofishonfriday:
+revert adding actions "Xenakios/SWS/NF: Set selected tracks heights to A/B, respect height locked tracks" (added in v2.12.0pre)
+refactor "Xenakios/SWS: Set selected tracks heights to A/B" (use COMMAND_T::user)

fixes #1221


nofishonfriday:
continuation/update of https://github.com/reaper-oss/sws/pull/1346 (which I can't reopen because I've force-pushed to it in 'closed' state).

I think the zoomed track(s) height calculation in `VertZoomRange()` would need updating too when called with `bMinimizeOthers == true` (because with `obeyHeightLock == true` it can now be off as the height locked tracks aren't minimized anymore which isn't accounted for in the current calculation ) but this is beyond my understanding of the code. :/

nofishonfriday:
@cfillion 
I've noticed a (new) problem with `SWS: Vertical zoom to selected tracks, minimize others ` with he new code from this PR:
If the action does nothing (this can happen now e.g. when all tracks are heightlocked), it seems to get added as a `ReentrantCmd` and thus eaten by `hookCommandProc()` hence it does nothing anymore on this and consecutive runs.
Any idea how to solve this?

sws-dbg

Zoom.cpp
```diff
@@ -108,11 +109,31 @@ void VertZoomRange(int iFirst, int iNum, bool* bZoomed, bool bMinimizeOthers, bo
...
+		}
+		else 
+		{
+			for (int i = 0; i <= GetNumTracks(); i++) {
+				MediaTrack* tr = CSurf_TrackFromID(i, false);
+				// NF: can't use SetMediaTrackInfo_Value(tr, "I_HEIGHTOVERRIDE", 0) here, because it would disable height lock
+				// https://forum.cockos.com/showthread.php?p=2202082#post2202082, #1221
+				SetMediaTrackInfo_Value(tr, "I_HEIGHTOVERRIDE", minTrackHeight);
+			}
+		}
```
cfillion, line 135: Somehow (I don't understand how the whole function works, perhaps it happens via action 40112?), setting the height override of non-locked tracks to `minTrackHeight` rather than 0 sometimes prevents them from being minimized:

Screencap

This fixes it (I also merged the two branches of the `obeyHeightLock` condition to simplify the code):

```suggestion
		for (int i = 0; i <= GetNumTracks(); i++)
		{
			MediaTrack* tr = CSurf_TrackFromID(i, false);
			const bool locked = SWS_IsTrackHeightLocked(tr);

			if (!obeyHeightLock || !locked)
				SetMediaTrackInfo_Value(tr, "I_HEIGHTOVERRIDE", locked ? minTrackHeight : 0);
		}
```

Zoom.cpp
```diff
@@ -124,7 +145,7 @@ void VertZoomRange(int iFirst, int iNum, bool* bZoomed, bool bMinimizeOthers, bo
 		for (int i = 0; i < iNum; i++)
 		{
 			MediaTrack* tr = CSurf_TrackFromID(i+iFirst, false);
-			if (bZoomed[i])
+			if ((obeyHeightLock && bZoomed[i] && !SWS_IsTrackHeightLocked(tr) || (!obeyHeightLock && bZoomed[i])))
```
cfillion, line 148:
```
../Zoom.cpp:148:38: warning: '&&' within '||' [-Wlogical-op-parentheses]
                        if ((obeyHeightLock && bZoomed[i] && !SWS_IsTrackHeightLocked(tr) || (!obeyHeightLock && bZoomed[i])))
```

```suggestion
			if (bZoomed[i] && (!obeyHeightLock || !SWS_IsTrackHeightLocked(tr))
```

Zoom.cpp
```diff
@@ -243,8 +264,23 @@ void VertZoomRange(int iFirst, int iNum, bool* bZoomed, bool bMinimizeOthers, bo
...
+				MediaTrack* tr = CSurf_TrackFromID(i, false);
+				if (!SWS_IsTrackHeightLocked(tr))
+					GetSetMediaTrackInfo(CSurf_TrackFromID(i, false), "I_HEIGHTOVERRIDE", &trackHeight);
+			}
+		}
+		else
+		{
+			for (int i = 0; i <= GetNumTracks(); i++)
+			{
+				MediaTrack* tr = CSurf_TrackFromID(i, false);
```
cfillion, line 280: This `tr` is unused. Also, the two branches of the condition could be merged.

```
../Zoom.cpp:281:17: warning: unused variable 'tr' [-Wunused-variable]
```

sws_util.cpp
```diff
@@ -542,3 +542,8 @@ void UpdateStretchMarkersAfterSetTakeStartOffset(MediaItem_Take* take, double ta
 	}
 	UpdateItemInProject(GetMediaItemTake_Item(take));
 }
+
+bool SWS_IsTrackHeightLocked(MediaTrack* track)
+{
+	return GetMediaTrackInfo_Value(track, "B_HEIGHTLOCK") ? true : false;
+}
```
cfillion, line 549: Missing newline at end of file and the ternary operator is not redundant here (1/0 is already true/false).

cfillion (commented):
> it can now be off as the height locked tracks aren't minimized anymore

What are the repro steps for this?

> If the action does nothing (this can happen now e.g. when all tracks are heightlocked) [...] hence it does nothing anymore on this and consecutive runs.

I think that's the intended behavior, since it's a toggle action (it works if running the action twice at the last step). Ideally `SaveZoomSlice` should not be called if nothing was done. EDIT: Along with `g_bASToggled = true` and `g_togAS.Get()->Save(true, iType < 3)` in `TogZoom`.

Xenakios/MixerActions.cpp
```diff
@@ -292,15 +297,32 @@ void DoStoreSelTraxHeights(COMMAND_T*)
...
-		MediaTrack* CurTrack = CSurf_TrackFromID(i+1,false);
-		if (*(int*)GetSetMediaTrackInfo(CurTrack,"I_SELECTED",NULL))
+		for (int i = 0; i < GetNumTracks(); i++)
 		{
-			GUID curGUID=*(GUID*)GetSetMediaTrackInfo(CurTrack,"GUID",NULL);
-			for (int j=0;j<(int)g_vec_trackheighs.size();j++)
-				if (GuidsEqual(&curGUID, &g_vec_trackheighs[j].guid))
-					GetSetMediaTrackInfo(CurTrack,"I_HEIGHTOVERRIDE",&g_vec_trackheighs[j].Heigth);	
+			MediaTrack* CurTrack = CSurf_TrackFromID(i + 1, false);
+			if (*(int*)GetSetMediaTrackInfo(CurTrack, "I_SELECTED", NULL) && !!GetMediaTrackInfo_Value(CurTrack, "B_HEIGHTLOCK"))
```
cfillion, line 305: This changes height-locked tracks instead of the opposite.

The two branches of the condition can be merged:

```cpp
void DoRecallSelectedTrackHeights(COMMAND_T*)
{
	for (int i = 0; i < GetNumTracks(); i++)
	{
		MediaTrack* CurTrack = CSurf_TrackFromID(i + 1, false);
		if (!GetMediaTrackInfo_Value(CurTrack, "I_SELECTED") || (NF_IsObeyTrackHeightLockEnabled() && GetMediaTrackInfo_Value(CurTrack, "B_HEIGHTLOCK")))
			continue;

		GUID curGUID = *(GUID*)GetSetMediaTrackInfo(CurTrack, "GUID", NULL);
		for (int j = 0; j < (int)g_vec_trackheighs.size(); j++)
			if (GuidsEqual(&curGUID, &g_vec_trackheighs[j].guid))
				GetSetMediaTrackInfo(CurTrack, "I_HEIGHTOVERRIDE", &g_vec_trackheighs[j].Heigth);
	}
```

Shouldn't `DoStoreSelTraxHeights` be also updated to match?

nofishonfriday:
updated.
If you find something that should be changed to make it mergeable feel free to commit directly if you want to. :)

cfillion:
I've split the commits for you here: https://github.com/reaper-oss/sws/compare/master...cfillion:track-height-split-commits.

I added a few additional changes on top:

- Bugfix commit: restored the original behavior of setting the height to 0 for non-height locked tracks
- Refactor commit:
  - replaced `TrackHeightA` and `TrackHeightB` with a `TrackHeight` array
  - simplified toggling of `g_CurTrackHeightIdx`
- Obey lock height commit: simplified storing the bool into the .ini by replacing the `sprintf` call

cfillion:
Ah, I think I see what you meant earlier by "height locked tracks aren't minimized anymore which isn't accounted for in the current calculation":

Screenshot

Potential fix:

```diff
diff --git a/Zoom.cpp b/Zoom.cpp
index 4b005425..5a18b615 100644
--- a/Zoom.cpp
+++ b/Zoom.cpp
@@ -238,7 +238,11 @@ void VertZoomRange(int iFirst, int iNum, bool* bZoomed, bool bMinimizeOthers, bo
 				MediaTrack* tr = CSurf_TrackFromID(i+iFirst, false);
 				if (TcpVis(tr))
 				{
-					trackHeight = GetTrackHeightFromVZoomIndex(tr, iZoom);
+					const bool locked = GetMediaTrackInfo_Value(tr, "B_HEIGHTLOCK");
+					if (obeyHeightLock && locked)
+						trackHeight = GetMediaTrackInfo_Value(tr, "I_HEIGHTOVERRIDE");
+					else
+						trackHeight = GetTrackHeightFromVZoomIndex(tr, iZoom);
 
 					int envHeight = 0;
 					if (i < (int)envelopeHeights.size())
```

nofishonfriday:
> Travis CI - Pull Request Failing after 9m — Build Failed 
`<array>` is C++11, right?
https://stackoverflow.com/questions/12516830/mac-osx-10-7-4-xcode-4-4-1-no-array-header-file

Then should I revert to the old version?
Also I now remember why I tend to stuff several things in one commit: :dagger: D
The revert should actually be included in the Refactor commit, but since it's not the latest commit anymore I don't know how to modify/amend it.

cfillion:
Right, it's C++11 (oops!). It should be an old-school C array: `int TrackHeight[2];` in the header.

You can use git rebase's interactive mode to amend a commit prior the last (if you use a GUI tool it probably has a similar feature):

```sh
git rebase -i HEAD~4
# replace "pick" with "edit" on the desired commit(s) to modify, save&quit the editor
# do the change
git add Xenakios/Parameters.h
git commit --amend
git rebase --continue
```

(HEAD~4 to review the last four commits, you can put a branch name there to review all commits since it: `git rebase -i master`)

https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History

nofishonfriday:
> Potential fix (needs testing):
Looks good to me (when admittedly briefly tested), added (https://github.com/reaper-oss/sws/commit/524d71128c007c8f383c88b9018e09c3b3d2e58e).

Are we done now already for 2-12-1? :P
Big thanks for your help as always!

cfillion:
> Potential fix (needs testing):

The "minimize others" actions have the same bug (https://i.imgur.com/1IZwiNu.gif). Same fix:

```diff
diff --git a/Zoom.cpp b/Zoom.cpp
index 5a18b615..3cf67f75 100644
--- a/Zoom.cpp
+++ b/Zoom.cpp
@@ -144,7 +144,14 @@ void VertZoomRange(int iFirst, int iNum, bool* bZoomed, bool bMinimizeOthers, bo
 			{
 				if (TcpVis(tr))
 				{
-					int trackHeight = GetTrackHeightFromVZoomIndex(tr, 0);
+					const bool locked = GetMediaTrackInfo_Value(tr, "B_HEIGHTLOCK");
+
+					int trackHeight;
+					if (obeyHeightLock && locked)
+						trackHeight = GetMediaTrackInfo_Value(tr, "I_HEIGHTOVERRIDE");
+					else
+						trackHeight = GetTrackHeightFromVZoomIndex(tr, 0);
+
 					trackHeight += CountTrackEnvelopePanels(tr) * GetEnvHeightFromTrackHeight(trackHeight);
 					iNotZoomedSize += trackHeight;
 				}

```

---

Nitpick: "SWS/NF: Toggle track height - and vertical zoom actions obey track height lock". When reading that quickly, it kinda sounds like it's going to "Toggle track height", while in fact it toggles "obey track height lock". How about naming it something like "SWS/NF: Toggle obey track height lock in vertical zoom and track height actions"?

cfillion (approved):

cfillion:
Merged into `sws-2-12-1`! I added a change to the last commit for avoiding these two MSVC warnings:

```
..\Zoom.cpp(153): warning C4244: '=': conversion from 'double' to 'int', possible loss of data
..\Zoom.cpp(252): warning C4244: '=': conversion from 'double' to 'int', possible loss of data
```

PR #1435, opened 2020-10-28
Allow start and/or end to be ignored in TrimItem_UseNativeTrimActions.

### LINKED ISSUE #1114: FR: BR_SetItemEdges -1 value for not moving edge
Hi !

It would be nice if we could pass "-1" (or any negative number) as parameter value into BR_SetItemEdges, to ignore one edge 😄 

Thx !

jameadows:
If either the 'start' or 'end' parameters are less than zero the original
start/end value is kept.

closes #1114

### DIFF
--- a/Breeder/BR_Util.cpp
+++ b/Breeder/BR_Util.cpp
@@ -1078,6 +1078,22 @@ bool TrimItem_UseNativeTrimActions(MediaItem* item, double start, double end, bo
 	if (!item)
 		return false;
 
+	// Fetch the current start and length for the item
+	double itemPos = GetMediaItemInfo_Value(item, "D_POSITION");
+	double itemLen = GetMediaItemInfo_Value(item, "D_LENGTH");
+
+	// If start is negative then keep the current start position
+	if(start < 0)
+	{
+		start = itemPos;
+	}
+
+	// If end is negative then keep the current end position
+	if(end < 0)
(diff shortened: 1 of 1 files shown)


jameadows:
Tested in action editor like this, using various values for start/end parameters.
```
item=GetSelectedMediaItem(0,0);
extension_api("BR_SetItemEdges",item, -1, 32);
```

Breeder/BR_Util.cpp
```diff
@@ -1078,6 +1078,22 @@ bool TrimItem_UseNativeTrimActions(MediaItem* item, double start, double end, bo
...
+	if(start < 0)
+	{
+		start = itemPos;
+	}
+
+	// If end is negative then keep the current end position
+	if(end < 0)
+	{
+		end = itemPos + itemLen;
+	}
```
cfillion, line 1095: Nice feature!

There was already a check for `if (start < 0)` a few lines below, which is now redundant (should be removed).

A small improvement suggestion to match the style of the surrounding code:
```suggestion
	if (start < 0)
		start = itemPos;

	// If end is negative then keep the current end position
	if (end < 0)
		end = itemPos + itemLen;
```

jameadows: Good catch!  Thanks for the review, I've pushed your requested changes.

PR #1437, opened 2020-10-31
Allow start and/or end to be ignored in TrimItem_UseNativeTrimActions

### LINKED ISSUE #1114: FR: BR_SetItemEdges -1 value for not moving edge

jameadows:
If either the 'start' or 'end' parameters are less than zero the original start/end value is kept.

Co-authored-by: Christian Fillion <cfillion@users.noreply.github.com>

closes #1114

### DIFF
--- a/Breeder/BR_Util.cpp
+++ b/Breeder/BR_Util.cpp
@@ -1078,17 +1078,25 @@ bool TrimItem_UseNativeTrimActions(MediaItem* item, double start, double end, bo
 	if (!item)
 		return false;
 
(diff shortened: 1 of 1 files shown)


jameadows:
Hi, I redid this pull request b/c I ended up with five commits, didn't realize until I'd pushed.  This changeset includes the changes requested in review.  Thanks!

PR #1432, opened 2020-10-22
Version 2.12.1.1

cfillion:
~Preparing a hotfix update to 2.12.1.~ ~EDIT: Promoting that to a full bugfix update given the recent PRs.~ EDIT2: Reverting back to hotfix (+ trivial PRs) for quicker stable adoption.

---

### v2.12.1.1 pre-release build

Accessibility:

+ Support OSARA's shortcut help mode in cycle, continuous and MIDI editor actions

macOS:

+ Add macOS ARM64 build
+ Fix NF Eraser tool cursor not appearing (report here)

MIDI:

+ Fix MIDI CC shape loss after running any of the following actions (REAPER v6+ only, issue 1466):
    - SWS/BR: Delete tempo marker and preserve position and length of items (including MIDI events)
    - SWS/BR: Delete tempo marker and preserve position and length of selected items (including MIDI events)
    - SWS/BR: Enable "Ignore project tempo" for selected MIDI items preserving time position of MIDI events (use tempo at item's start)
+ Fix FNG MIDI actions treating only the first selected item in REAPER v6 (regression from v2.12.1, issue 1429)

Miscellaneous:

+ Fix SWS/AW: Paste not setting the time selection when "Trim contents behind media items when editing" is enabled (regression from v2.11)
+ Obey the "Move edit cursor when pasting/inserting media" preference in SWS/AW: Paste even when "Trim contents behind media items when editing" is disabled

ReaScript API:

+ Deprecate BR_GetMediaTrackGUID (use GetSetMediaTrackInfo_String from REAPER instead)
+ Deprecate ULT_GetMediaItemNote and ULT_SetMediaItemNote (use GetSetMediaItemInfo_String instead)
+ Fix fetching tracks from background project tabs using BR_GetMediaTrackByGUID
+ Fix generation of sws_python64.py
+ Preserve existing edges when passing negative values to BR_SetItemEdges (issue 1114)
+ Prevent NF_AnalyzeTakeLoudness's progress popup from stealing focus from other apps (report here)
+ Remove NF_ReadID3v2Tag (use GetMediaFileMetadata from REAPER instead)

PR #1444, opened 2020-11-03
SWS/S&M: Toggle selected FX bypass for selected tracks

### LINKED ISSUE #1443: "SWS/S&M: Toggle selected FX bypass for selected tracks" action only toggles first selected FX, not all selected
With more than one effect selected in the track FX UI, using the menu command to bypass will in fact bypass all selected effects. This action, however, will only bypass the first effect selected.

jameadows:
Iterate through selected effects on selected tracks and toggle
the enabled state.

Fixes #1443


jameadows:
That last push falls back to the legacy behavior if the FX window is not open for a selected track, i.e. it toggles the selected FX index for the track(s) if it's a valid index.

SnM/SnM_FX.cpp
```diff
@@ -101,14 +102,37 @@ bool SetOrToggleFXBypassSelTracks(const char* _undoMsg, int _mode, int _fxCmdId,
...
+								// toggle enabled state
+								TrackFX_SetEnabled(tr, index, !TrackFX_GetEnabled(tr, index));
+								updated = true;
+							}
 						}
+						else if(fxId < fxcnt)
+						{
+							int index=-1;
+							// enumerate selected FX
+							while( (index = CF_EnumSelectedFX(fxChain, index)) >= 0)
```
cfillion, line 126: `fxChain` is NULL at this point, so the action fails to find the last active FX when the chain is closed. The last active FX index is stored in `fxId`.

jameadows: > `fxChain` is NULL at this point, so the action fails to find the last active FX when the chain is closed. The last active FX index is stored in `fxId`.
> 
Strange, I have the same code in both blocks, but I remember testing this with FX window opened and closed.  Anyhow, sorry, I pushed an update that reverts to the previous behavior if the FX window is not opened.

> Updating the meaning of "selected FX" from "the last FX shown in the chain" to "all selected FXs in the chain" is interesting, but I think it should be done across all SWS actions dealing with "selected FX". [#1432 (comment)](https://github.com/reaper-oss/sws/pull/1432#issuecomment-739555373)

IIUC the original request was to make the action behave the same as the menu command (Ctrl-B).  I searched for other actions that affected selected FX, are these what you were wanting to include?:

SWS/S&M: Show FX chain for selected tracks (selected FX)
SWS/S&M: Float selected FX for selected tracks
SWS/S&M: Unfloat selected FX for selected tracks
SWS/S&M: Toggle float selected FX for selected tracks
SWS/S&M: Move selected FX up in chain for selected tracks
SWS/S&M: Move selected FX down in chain for selected tracks
SWS/S&M: Remove selected FX for selected tracks
SWS/S&M: Toggle selected FX online/offline for selected tracks
SWS/S&M: Set selected FX online for selected tracks
SWS/S&M: Set selected FX offline for selected tracks
SWS/S&M: Bypass selected FX for selected tracks
SWS/S&M: Unbypass selected FX for selected tracks
SWS/S&M: Trigger preset for selected FX of selected track (MIDI/OSC only)
SWS/S&M: Trigger next preset for selected FX of selected tracks
SWS/S&M: Trigger previous preset for selected FX of selected tracks
SWS/S&M: Reassign MIDI learned channels of selected FX for selected tracks (prompt)

daeavelwyn:
Hi folks,
Reaper 6.24rc7 here on windows 64 and sws v2.12.1.3 featured build (February 6, 2021) and this action still doesn't work. I get the same issue as the one mentioned in #1443 
any workaround ?

PR #1445, opened 2020-11-03
AW: make potentially altered config variables persistent

### LINKED ISSUE #1442: command ID _SWS_AWCLRTIMESELCLKTOG (SWS/AW: Toggle 'link time selection and edit cursor') has no persistent state
Originally reported on REAPER forums:
https://forum.cockos.com/showthread.php?p=2360154

The _SWS_AWCLRTIMESELCLKTOG action will effectively change the following preferences:
Preferences > Editing behaviour > Move edit cursor to start of time selection on time selection change' and
'[same path] > Clear time selection when edit cursor moves on click in arrange view'

However, the setting does not persist (not written to reaper.ini) as when changing the affected preferences directly via the REAPER Preferences UI.

nofishonfriday:
closes #1442

### DIFF
--- a/Misc/Adam.cpp
+++ b/Misc/Adam.cpp
@@ -3354,3 +3354,24 @@ int AdamInit()
 
 	return 1;
 }
+
+void WriteAlteredConfigVarsToIni()
+{
+	const char* alteredConfigVars[] = {
+		"projmetroen",
+		"relativeedges",
+		"itemclickmovecurs",
+		"itemtimelock"
+	};
+
+	char tmp[256];
+	for (const auto& configVar : alteredConfigVars) {
+		snprintf(tmp, sizeof(tmp), "%d", *ConfigVar<int>(configVar));
+		WritePrivateProfileString("reaper", configVar, tmp, get_ini_file());
+	}
+}
+
+void AdamExit()
+{
+	WriteAlteredConfigVarsToIni();
+}
--- a/Misc/Adam.h
+++ b/Misc/Adam.h
@@ -28,6 +28,7 @@
 #pragma once
 
 int AdamInit();
+void AdamExit();
 void AWFillGapsAdv(const char* title, char* retVals);
 void UpdateGridToolbar();
 void UpdateTrackTimebaseToolbar();
--- a/sws_extension.cpp
+++ b/sws_extension.cpp
@@ -660,6 +660,7 @@ error:
 				PadreExit();
 				SNM_Exit();
 				BR_Exit();
+				AdamExit();
 			}
 			return 0; // makes REAPER unloading us
 		}


Misc/Adam.cpp
```diff
@@ -3354,3 +3354,24 @@ int AdamInit()
 
 	return 1;
 }
+
+void WriteAlteredConfigVarsToIni()
+{
+	const char* alteredConfigVars[] = {
+		"projmetroen",
```
cfillion, line 3361: Same as `itemtimelock` below: this changes the user's defaults for new projects.

Misc/Adam.cpp
```diff
@@ -3354,3 +3354,24 @@ int AdamInit()
...
+		"relativeedges",
```
cfillion, line 3362: `relativeedges` is never modified by SWS: the `AWRelEdge*` actions were removed in 1ad238265873b71faa14b4e0821b5e901c1b245b. Those unused functions should probably be deleted...

Misc/Adam.cpp
```diff
@@ -3354,3 +3354,24 @@ int AdamInit()
...
+		"itemclickmovecurs",
+		"itemtimelock"
```
cfillion, line 3364: This saves the current project timebase as the default for new projects without asking. `itemtimelock` is a project configvar.

nofishonfriday: Thanks. Hm...so the only one that should be made persistant is `itemclickmovecurs`, right?
Should I then rather revert the `WriteAlteredConfigVarsToIni()` and write to ini in the functions itself which alter it? (7 functions if I counted correctly.) This would also prevent the writing to ini at exit when actually nothing has changed. On the other hand, it would write every time when toggled and not just one time at exit.

cfillion: > Should I then rather revert the WriteAlteredConfigVarsToIni() and write to ini in the functions itself which alter it? (7 functions if I counted correctly.)

I think a `PersistItemClickMoveCurs` called from the 7 actions makes sense. Many of REAPER's own options set/toggle actions immediately write to the .ini.

Writing on exit might conflict with REAPER or another extension/script wanting to save a conflicting value (possibly from user interaction made after those SWS actions were used).

cfillion: If going the write immediately route, can you cherry-pick fb46dd66e15ec06032dbcb95be4647019f2967b8 first and add the persist fix on top? (Then I'll close #1446 in favor of this PR.)

nofishonfriday: I'm AFK now for a couple of hours. If you want to merge this in 2.12.1.1 fast (?), don't hesitate to go ahead and do it.
Otherwise I'd try when I'm back. (I like the idea of adding a `.persist()` to `configvar.h`, if I understood the idea correctly.)

cfillion: There's no hurry, but I have some free time now, so I'll give it a go.

nofishonfriday:
closing because superseded by #1446

PR #1447, opened 2020-11-04
    Increase 'save time selection' slots from five to sixteen.

### LINKED ISSUE #1405: We need more "save time selection" slots please :)
Hey :) 

Please consider adding more slots for time selection saving actions. I know nothing about coding, but I suspect it's not a very hard thing to implement, I suspect also that the current limitations, 5, has been set "arbitrary". I can promise you there are use cases where you need more of them. 10 would be cool. 16 (just like cursors) would be splendid... 

Thanks ! Love SWS :)

jameadows:
Fixes #1405


jameadows:
Is there a 'vetting' process for which feature requests to include?  I'm working on issues to get myself up to speed on the code, picking stuff that looks doable, but I don't want to waste anyone's time having to review PR's for things that aren't in line with the project's goals.

cfillion:
Increasing the quantity of actions merely postpones the problem to later. These actions should be converted to one dynamic action, whose slot quantity is user-configurable via S&M.ini. See #1364 for a recent example of such a conversion.

> I don't want to waste anyone's time having to review PR's for things that aren't in line with the project's goals.

Fixing things that aren't quite right for users (like #1405) is very well in line with making SWS better. 👍

jameadows:
> Increasing the quantity of actions merely postpones the problem to later. These actions should be converted to dynamic actions, whose slot quantity is user-configurable via S&M.ini. See #1364 for a recent example of such a conversion.
> 
Thanks, it did seem a bit cumbersome to do it the way I did.  I'll follow up with your suggestion.

> Fixing things that aren't quite right for users (like #1405) is very well in line with making SWS better. +1

:+1:

nofishonfriday:
@jameadows 
As a sidenote, what's a little different to the example given by @cfillion (#1364) is that the time selction slots are saved in the project file on project save for persistency. So this may be something to keep in mind when converting to a dynamic action (but I haven't really looked into it closely). 
time_sel_slots

jameadows:
This latest change converts save/restore time selection to a dynamic command, default 16 slots.  I also converted save/restore loop selection since it basically came for free.  This also adds undo support so the project gets marked as modified when these actions run.  The settings are correctly persisted and restored from the project file.

One of the checks failed (Linux x86 32 bit) but it looks like a test environment problem.  Is there a way to rerun the test to see if was just a transient failure?

PS: Thanks for the suggestions and pointers!

nofishonfriday:
>  Is there a way to rerun the test to see if was just a transient failure?

Maybe you can push force your latest commit again to trigger a restart of the build process.
Never tried this myself though, but I know push forcing a new commit does trigger a rebuild.

> This also adds undo support so the project gets marked as modified when these actions run. 

Hm..I'm not sure it needs creating a seperate undo entry in the undo history for these actions, in my comment above about marking the project dirty I rather though about just calling [MarkProjectDirty()](https://www.extremraym.com/cloud/reascript-doc/#MarkProjectDirty).
But let's see what @cfillion thinks.

jameadows:
An empty push did cause it to run the validation tests again, this time they passed.  The first time (when it failed) there was some issue with apt, seemed to fail before it even got to testing the code.

Let me know what you guys think about undo vs just marking the project as dirty, or for that matter any other suggestions that come to mind.  Thanks!

cfillion:
> Hm..I'm not sure it needs creating a seperate undo entry in the undo history for these actions, in my comment above about marking the project dirty I rather thought about just calling MarkProjectDirty() 

Agreed. Using `MarkProjectDirty` would make it consistent with other SWS actions such as "SWS/BR: Save edit cursor position".

> One of the checks failed (Linux x86 32 bit) but it looks like a test environment problem. Is there a way to rerun the test to see if was just a transient failure?

Random build failure happens from time to time. I can manually retrigger the failed jobs when it happens (no need to re-push).

nofishonfriday:
@jameadows
Thanks. I think the currently 5 commits could be squashed down to two (one for conversion to dyn. actions, one for marking project dirty), for a little cleaner SWS commit history.

Quoting from the [Submitting pull requests](https://github.com/reaper-oss/sws/wiki/Submitting-pull-requests) Wiki entry:
> Don't hesitate to squash commits to keep the history clean. Amending commits and rewriting branch history is fine until the PR is merged.

jameadows:
> @nofishonfriday 
> Thanks. I think the currently five commits could be squashed down to two (one for conversion to dyn. actions, one for marking project dirty), for a little cleaner SWS commit history.
> 
Hope I got that right.

nofishonfriday:
Now the number of commits has increased to eight, instead of being reduced, so it's not quite there yet. :)
To elaborate on my above comment:
Imo this pull request should contain only two commits, the content/commit messages being something like

- Convert save/restore time selection slot and save/restore loop selection slot to dynamic actions

Default number of slots is 16, max 32
Fixes #1405

- Mark project dirty when saving time/loop selection slots

Are you familiar with rewriting Git history (amending/squashing commits etc.), also how are you using Git, via command line or a graphical Git client?

jameadows:
> Now the number of commits has increased to eight, instead of being reduced, so it's not quite there yet. :)

D'oh!

> Are you familiar with rewriting Git history (amending/squashing commits etc.), also how are you using Git, via command line or a graphical Git client?
> 
I'm using command line Git.  I think what I'm missing is how to merge multiple commits once they've been pushed to my working branch.

> (An alternative way would be to close this one and re-submit a 'cleaned up' version.)
>

I'll do this.
 
> edit:
> To explain the reasoning behind it:

That makes total sense, no need to see all the work in progress and thrashing, just the important milestones.

Thanks!

nofishonfriday:
>  I think what I'm missing is how to merge multiple commits once they've been pushed to my working branch.

I admit I'm using a Git client (SourceTree) where it can be done in a graphical way, so I'm not the person to advice on Git command line usage. :D
But basically it's [squashing commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) locally and then do a `git push --force`.

jameadows:
Closing this PR, have created a new one with the same changes but without all the extra commits

jameadows:
> I admit I'm using a Git client (SourceTree) where it can be done in a graphical way, so I'm not the person to advice on Git command line usage. :D
> But basically it's [squashing commits](https://git-scm.com/book/en/v2/Git-Tools-Rewriting-History) locally and then do a `git push --force`.

Thanks, I know just enough about Git to be dangerous ;)  I think I'll create a sandbox repo where I can trying things out without doing any real damage.  

Cheers,
Joe

cfillion:
> I'm using command line Git. I think what I'm missing is how to merge multiple commits once they've been pushed to my working branch.

```sh
git rebase -i origin/master
```

This will open a text editor: edit the commands (the first word) to be fixup (or squash) for all except the first commit. This will merge all commits into the first one. You can use the reword command in the first line to update the first commit's message (via another editor instance).

Check the history locally before pushing (add -p to view the diffs):
```sh
git log --oneline --graph origin/master..HEAD
```
Then push the new branch history:
```sh
git push --force
```

jameadows:
> I recommend using C++11-style `nullptr` instead of `NULL` for new code.
> 
...
> 
> ```shell
> git push --force
> ```

Great info, thanks.  I'll try this out with the NULL -> nullptr change.

nofishonfriday:
@jameadows 
I wanted to elaborate again a little on the number of commits per PR, because I remember it took myself some time to get right when starting out.
Just in case it came across this way, it's not always one PR = one commit, it's more like (ideally) one 'atomic thing changed' (still not quite sure of the accurate expression) = one commit.
[This](https://github.com/reaper-oss/sws/pull/1453) PR by cfillion is a good example, where he has splitted it into four commits.

Hope that helps. :)

PR #1449, opened 2020-11-05
update BR grid calculation for swing grid

### LINKED ISSUE #1137: reaper.BR_GetNextGridDivision() does not take swing into account
Unfortunately, reaper.BR_GetNextGridDivision() does not take swing into account, when set.
Is it possible to include it?

nofishonfriday:
fixes #1137

Needs testing as this function is used in various places:
BR grid vs swing

Initial test seems ok:
BR grid vs swing


nofishonfriday:
Sorry, this is obviously over my head to do it correctly. :/

PR #1458, opened 2020-11-16
BR_GetMediaTrackByGUID: support any project tab (not just the current one)

nofishonfriday:
reported by X-Raym via Slack:
https://app.slack.com/client/T0HQ3KQUQ/C0XA86FML/thread/C0XA86FML-1605548134.190000

`BR_GetMediaTrackGUID()` doesn't seem to work for tracks in background projects either (as doesn't native `GetTrackGUID()`), but it seems to be a REAPER API bug, as it doesn't seem to evaluate the track pointer as valid (see pic).
I've added a note in the API doc, we could also (additionally) mark it as deprecated (in favour of the native `GetTrackGUID()`?

BR_GetMediaTrackGUID multitabs

````
reaper.Undo_BeginBlock()
proj = reaper.EnumProjects(1) -- second tab
t = reaper.GetTrack(proj, 3)
r, n = reaper.GetTrackName( t )
validate = reaper.ValidatePtr2(proj,t, 'MediaTrack*')
GUID_GetTrackGUID = reaper.GetTrackGUID(t)
GUID = reaper.BR_GetMediaTrackGUID (t)
retval, GUID_MediaTrackInfo = reaper.GetSetMediaTrackInfo_String( t, "GUID", '', false )

a_track_by_GUID = reaper.BR_GetMediaTrackByGUID( proj, GUID )

reaper.Undo_EndBlock("a", -1)
````

### CODE REVIEW

**Comment by cfillion** [comment]
> BR_GetMediaTrackGUID() doesn't seem to work for tracks in background projects either (as doesn't native GetTrackGUID()), but it seems to be a REAPER API bug

This comes from the `ValidatePtr2` check at sws_utils.cpp:280. `BR_GetMediaTrackGUID` could be fixed by passing the track's project (P_PROJECT) to `TrackToGuid`.

**Comment by nofishonfriday** [comment]
> BR_GetMediaTrackGUID could be fixed by passing the track's project (P_PROJECT) to TrackToGuid.

That's a nice idea, thanks, but - `GetMediaTrackInfo_Value()` returns a double, how do I convert it to a ReaProject (no successs so far what I tried)?

**Comment by cfillion** [comment]
```cpp
ReaProject *proj = static_cast<ReaProject *>(GetSetMediaTrackInfo(track, "P_PROJECT", nullptr));
```

This would fix the function itself not supporting tracks from background projects. The REAPER-side validation issue of "expected MediaTrack*" will remain though (I didn't notice that before).

In my opinion this function should be deprecated in favor of `GetSetMediaTrackInfo_String`.

X-Raym:
Thanks for your work on this !

I second the use of GetSetMediaTrackInfo_String, GetTrackGUID doesn't work with other project tab.

For info, the BR_GetMediaItemByGUID function seems to works.

Issue #1404, opened 2020-09-03
Region playlist does not work with VoiceOver accessibility on MacOS


mcclure:
A sightless REAPER user on Reddit is reporting the Region Playlist pane in Reaper SWS does not work with the MacOS "VoiceOver" feature. He says VoiceOver works fine with the rest of REAPER and SWS. It sounds like this is blocking work for him.

https://www.reddit.com/r/synthesizers/comments/ikwe10/completely_off_topic_but_im_desperate_looking_for/

Can someone look at this? I found an SnM_RegionPlaylist.cpp in the source and it seems simple but without knowing how this "WDL" you are using handles accessibility it wasn't clear to me what needs to be fixed (I have a REAPER evaluation version but don't know how to use it…)

soundwarrior20:
Hello, I am the user who posted about this on reddit.  I now have more information about what I think needs to be done to improve accessibility of the region playlist. ,  please see the below forum thread  for this information I really hope somebody can help with this:  
https://forum.cockos.com/showthread.php?t=241888

nofishonfriday:
Hi @soundwarrior20,
just to let you know, I'm looking into this issue and let you know when there are news (though I can't promise anything :)) .
Please note that I'm doing this in my spare time so progress will probably be rather slow..
If you find a professional developer meanwhile who is up for hiring that's fine with me also of course. :)

mespotine:
Could this be an option or way for a Text2Speech-function, which we could use in Lua-scripts?

soundwarrior20:
@ mespotine there is a way to do this,look in the osara API look for the function osara_ _output message if its there you can simply give it a string to speek. this is what i've been told. also you can also look in to the speek function accessichords: 
https://github.com/Timtam/AccessiChords by @Timtam  hope this helps :) @nofishonfriday your work is much appreciated thank you very much :)

nofishonfriday:
@soundwarrior20 
I'm currently thinking about a suitable implementation for the new actions.
'Add all regions' is quite straightforward (actually I've already written some code for it), but how about e.g. setting name and loop count?
The regions playlist lists the added regions in a 'spreadsheet like' manner where each line represents a region with the aforementioned attributes.
The lines (aka regions) can be selected. (Once any region is selected, which currently works via mouse only afaik, the selection can be changed via cursor up/down keys.)
[...]

PR #1467, opened 2020-11-24
Region Playlist: accessibility improvements

nofishonfriday:
As per the conversation in #1404:

new actions (displaying error messages if necessary conditions not met):
- SWS/S&M: Region Playlist - Add playlist (Accessibilty)
- SWS/S&M: Region Playlist - Add all regions (Accessibilty)
- SWS/S&M: Region Playlist - Set name of selected region (Accessibilty)
- SWS/S&M: Region Playlist - Set loop count of selected region (Accessibility)
- SWS/S&M: Region Playlist - Display context menu (Accessibility) [not implemented]

new functionality:
- TAB selects first region (if present)

TODO:
- implement 'Display context menu' action
- remove '(Accessibilty)'  in action name?

I could use help with implementing the 'show context menu' (at mouse cursor) action.
I've added '(Accessibility)' to the new actions names for easier filtering/testing during development stage, could be removed for release.

### DIFF
--- a/SnM/SnM.cpp
+++ b/SnM/SnM.cpp
@@ -500,6 +500,12 @@ static COMMAND_T s_cmdTable[] =
 	{{ DEFACCEL, "SWS/S&M: Region Playlist - Options/Enable shuffle (only in Region Playlist)" }, "S&M_PLAYLIST_OPT_SHUFFLE_ON", SetPlaylistOptionShuffle, NULL, 0},
 	{ { DEFACCEL, "SWS/S&M: Region Playlist - Options/Disable shuffle (only in Region Playlist)" }, "S&M_PLAYLIST_OPT_SHUFFLE_OFF", SetPlaylistOptionShuffle, NULL, 1},
 	{ { DEFACCEL, "SWS/S&M: Region Playlist - Options/Toggle shuffle (only in Region Playlist)" }, "S&M_PLAYLIST_OPT_TGL_SHUFFLE", SetPlaylistOptionShuffle, NULL, -1, IsPlaylistOptionShuffle},
+	// Accessibility
+	{ { DEFACCEL, "SWS/S&M: Region Playlist - Add playlist (Accessibility)" }, "S&M_PLAYLIST_ADD_PLAYLIST", AddPlaylist, NULL, -1 },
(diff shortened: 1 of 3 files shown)


ScottChesworth:
Hey @nofishonfriday, seems the trail went cold on this. Are you still up for getting it over the line? If so, could you refresh the AppVeyor build? I can be around to help out with Windows testing now, also have a keen Mac user at the ready.

nofishonfriday:
closing this as I'm not really active with SWS dev anymore (np if someone wants to take it as basis for a new PR)

ScottChesworth:
@RDMurray, any chance you'd have time to pick this up?

RDMurray:
@ScottChesworth Okay, I'll try. First of all I am not sure that adding a bunch of actions to the main section that require space on the keymap is the right solution.

It seems that the only thing mac specific to this issue is that VO+Shift+m doesn't work, is that right?

The other problem, needing to double click to edit cells in the list view also exists on windows. The media explorer is the same in this regard, although edit metadata tag in the context menu makes it possible if a bit long winded.

I haven't looked at it on mac yet. Does VO shift m work in the media explorer? 

It would be nice if NVDA could reliably route the mouse to a cell after using ctrl+alt+arrows but it doesn't seem to work. 

What are your thoughts on how this should work, both in this window and in the media explorer?

CC @jcsteh

ScottChesworth:
> I am not sure that adding a bunch of actions to the main section that require space on the keymap is the right solution.
> 
Agreed. That was the request posted on Reddit and in the forum. No idea why. IMO it would be better to spend time tightening up keyboard access in the GUI. However, if action equivalents are already written and functional, keep those around, because now key map layers are a thing there might be some interesting possibilities around reshaping arrangements on the fly.

> It seems that the only thing mac specific to this issue is that VO+Shift+m doesn't work, is that right?
> 
I think all the other issues with keyboard access are cross platform yep.

RDMurray:
It would be easy to add context menu items to edit each field, like in the media explorer. Shortcut keys would also be possible. Should the shortcut keys be hardcoded, or should this window add it's own keyboard section for customizability? No other SWS windows have their own section afaik so I don't know if that would be accepted. It's also more work.

ScottChesworth:
Can't see why the window needs its own section here man. Hard code them, and if users do want to customize their keystrokes, they can use NOFish's actions. Thoughts?

PR #1489, opened 2020-12-24
ReaScript: CF_EnumMediaSourceCues() - optionally get if a source cue is a chapter marker

nofishonfriday:
> v6.19 - December 18 2020
>   + API: PCM_SOURCE_EXT_ENUMCUES_EX distinguishes between cues and chapter markers

Lua test code:
```
reaper.ClearConsole()

function msg(m)
  return reaper.ShowConsoleMsg(tostring(m) .. "\n")
end

item = reaper.GetSelectedMediaItem(0, 0)
take =  reaper.GetActiveTake(item)
src = reaper.GetMediaItemTake_Source(take)

local i = 0
while true do
  -- retval, time, endTime, isRegion, name = reaper.CF_EnumMediaSourceCues( src, i ) -- optional param not provided
  retval, time, endTime, isRegion, name, isChapter = reaper.CF_EnumMediaSourceCues(src, i)
  msg("retval. ".. retval)
  msg("name: " .. name)
  if isChapter ~= nil then 
    msg("isChapter: " .. tostring(isChapter)) 
  end
  msg("\n")
  if retval == 0 then
    break
  end

  i = i + 1
end
```

Audio files with chapter markers are e.g. [here](https://auphonic.com/blog/2013/07/03/chapter-marks-and-enhanced-podcasts/).

edit:
I'm not sure if I get this [comment](https://github.com/justinfrankel/reaper-sdk/blob/bab8de9516516b4be24ff3ec5374d77508493e7d/reaper-plugins/reaper_plugin.h#L397) correctly:
> &1=DEPRECATED caller must call Extended(PCM_SOURCE_EXT_ENUMCUES, -1, &cue, 0) when finished

edit2:
But I think this is only relevant when _setting_, so probably no need to call above here?
 
On a side note:
reaper-sdk (including `reaper_plugin.h`) [moved](https://github.com/justinfrankel/reaper-sdk) to GitHub recently, maybe we could add it as a submodule in future?

### DIFF
--- a/cfillion/cfillion.cpp
+++ b/cfillion/cfillion.cpp
@@ -349,7 +349,7 @@ bool CF_GetMediaSourceRPP(PCM_source *source, char *buf, const int bufSize)
 }
 
 int CF_EnumMediaSourceCues(PCM_source *source, const int index, double *time,
-  double *endTime, bool *isRegion, char *name, const int nameSize)
+  double *endTime, bool *isRegion, char *name, const int nameSize, bool *isChapter)
 {
   if(!source)
     return 0;
@@ -368,6 +368,9 @@ int CF_EnumMediaSourceCues(PCM_source *source, const int index, double *time,
   if(name && cue.m_name)
     snprintf(name, nameSize, "%s", cue.m_name);
 
+  if (isChapter)
+      *isChapter = cue.m_flags & 4;
+ 
   return add ? index + add : 0;
 }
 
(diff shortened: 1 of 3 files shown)


cfillion/cfillion.cpp
```diff
@@ -367,6 +367,9 @@ int CF_EnumMediaSourceCues(PCM_source *source, const int index, double *time,
   if(name && cue.m_name)
     snprintf(name, nameSize, "%s", cue.m_name);
 
+  if (isChapter && (cue.m_flags & 4))
+      *isChapter = true;
```
cfillion, line 371: `isChapter` won't be cleared if the user calls the function with it initialized to `true`:

```
reaper.CF_EnumMediaSourceCues(src, i, 0, 0, true, '', true) -- weird usage, but it's a valid call
```
```suggestion
  if (isChapter)
      *isChapter = cue.m_flags & 4;
```

> I'm not sure if I get this comment correctly:
> > &1=DEPRECATED caller must call Extended(PCM_SOURCE_EXT_ENUMCUES, -1, &cue, 0) when finished
>
> But I think this is only relevant when setting, so probably no need to call above here?

`&1` is for the deprecated `PCM_SOURCE_EXT_ENUMCUES`. It does not apply to the newer `PCM_SOURCE_EXT_ENUMCUESEX`.

> reaper-sdk (including reaper_plugin.h) moved to GitHub recently, maybe we could add it as a submodule in future?

Yeah. They also have reaper_plugin_functions.h there. If they keep it up-to-date it could mean new SWS devs won't have to generate it themselves anymore.

The reaper_plugin.h here was slightly modified though (include paths, 437d7d0b45d8dd460062d7419d9268d671dcc3c0). Using the upstream version might require a bit of tweaking.

(The sudden move of reaper_plugin.h out of Landoleet broke ReaPack's CI builds which depended on it... ☃️ )

nofishonfriday: Thanks, updated.
>**If** they keep it up-to-date

Yeah. :P

PR #1492, opened 2021-01-01
Rework 'Xenakios/SWS: Rename take source files...' and 'Xenakios/SWS: Rename takes and source files...'

### LINKED ISSUE #407: Rename takes and sources overwrites other files in folder
_From [sergioar...@gmail.com](https://code.google.com/u/114349142085426040966/) on November 01, 2011 23:41:28_

What steps will reproduce the problem? Rename an item using "Rename takes and source files" command. What is the expected output? What do you see instead? Renaming the items WITHOUT overwriting others with the same names. A prompt would be nice to ask if they should be overwritten. 

What version of SWS extension are you using? (Extensions -> About SWS...)
2.1.0 build#1 What is your Reaper version number and CPU architecture (e.g. Reaper 3.75 x64)? 4.10 x86 What operating system are you running?  (e.g. Windows XP/7, OSX 10.5.8 etc, 64bit?) W7

_Original issue: http://code.google.com/p/sws-extension/issues/detail?id=407_

### LINKED ISSUE #1498: Bug: "Xenakios/SWS: Rename takes and source files..." deletes file with the same name on MAC.
I had need to rename manually several items.
But by accident I named one take the same as the another. And the old source file has been deleted.
It works normally on PC. Just renames take, but not source.
On MAC it deletes file!!

nofishonfriday:
- warn when filename already exists
closes #407, closes #1498

- fix not working if take is a section or reversed
report: https://forum.cockos.com/showpost.php?p=2300847&postcount=2840
(Tested a bit and seems to work so far..)

- clean ups

- both actions contained quite a bit of identical code, so I tried to combine them into one


Xenakios/TakeRenaming.cpp
```diff
@@ -223,40 +165,39 @@ void DoRenameTakeAndSourceFileDialog(COMMAND_T* ct)
 					if (newfnsplit[2].compare(fnsplit[2])!=0)
 						newfilename.append(fnsplit[2]);
 
-					MoveFile(oldname.c_str(),newfilename.c_str());
+					if (!MoveFile(oldname.c_str(), newfilename.c_str()))
```
cfillion, line 168: This does not prompt if `newfilename` already exists on non-Windows, where `MoveFile` is an alias for `rename` (defined in swell-types.h:75).

This condition would only be entered if the move fails (eg. permission issue, directory not found...). Offlining first is only useful on Windows.

sws_util.cpp
```diff
@@ -512,3 +512,15 @@ void UpdateStretchMarkersAfterSetTakeStartOffset(MediaItem_Take* take, double ta
...
+const char* SWS_GetSourceFileName(PCM_source* src)
+{
+	if (!src->GetFileName()) {
+		src = src->GetSource();
+		if (src)
+			return src->GetFileName();
+		else
+			return nullptr;
+	}
+	return src->GetFileName();
```
cfillion, line 525:
```suggestion
	if (src && !src->GetFileName())
		src = src->GetSource();
	return src ? src->GetFileName() : nullptr;
```

Xenakios/TakeRenaming.cpp
```diff
@@ -27,26 +27,27 @@
...
-	int takesToRename;
+	size_t takesToRename;
 	int curTakeInx;
 	bool batchnaming;
 
-} t_renameparams;
+} ;
 
-t_renameparams g_renameparams;
+t_renameparams g_renameparams = {};
```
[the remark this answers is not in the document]
nofishonfriday: Hm..if I remove the `= {}` I get these warnings. Why?
sws_-_Microsoft_Visual_Studio_2021-01-26_16-51-53

Xenakios/TakeRenaming.cpp
nofishonfriday: > This does not prompt if newfilename already exists on non-Windows, where MoveFile is an alias for rename (defined in swell-types.h:75).
Using `FileOrDirExists()` from `SnM_Util.h` instead should be ok?

> Offlining first is only useful on Windows.
Tbh not sure what this implies. :)
For `PCM_source::SetFileName()` (which is called later on) it says in `reaper_plugin.h`:
> this will only be called when offline
so it needs to be set offline anyway?

cfillion: > Using FileOrDirExists() from SnM_Util.h instead should be ok?

Yes.

> PCM_source::SetFileName() [...] this will only be called when offline

Ah, I didn't know that. Then yes it does need to be set offline first.

Xenakios/TakeRenaming.cpp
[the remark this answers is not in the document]
cfillion: That comes from Visual Studio's static code analyzer rather than the compiler (Build + IntelliSense). No idea why it's triggered even though `g_renameparams` is global...

nofishonfriday:
Thanks, PR updated.
Though the commit history got a bit effed up as there are several commits that depend on each other so I couldn't rebase anymore without conflicts.

Xenakios/TakeRenaming.cpp
```diff
@@ -223,40 +165,39 @@ void DoRenameTakeAndSourceFileDialog(COMMAND_T* ct)
...
+					for (size_t j=0;j<alltakes.size();j++)
 					{
 						PCM_source *thesrc=(PCM_source*)GetSetMediaItemTakeInfo(alltakes[j],"P_SOURCE",0);
-						if (thesrc && thesrc->GetFileName() && strcmp(thesrc->GetType(),"SECTION")!=0)
+						if (thesrc)
 						{
-							string fname;
-							fname.assign(thesrc->GetFileName());
-							if (oldname.compare(fname)==0)
+							if (!strcmp(SWS_GetSourceFileName(thesrc), oldname.c_str()))
```
cfillion, line 183: These two conditions can be merged with `&&` to save an indentation level. Or two levels by reversing them:

```cpp
if (!thesrc || strcmp(...))
  continue;
```

Xenakios/TakeRenaming.cpp
```diff
@@ -223,40 +165,39 @@ void DoRenameTakeAndSourceFileDialog(COMMAND_T* ct)
 					if (newfnsplit[2].compare(fnsplit[2])!=0)
 						newfilename.append(fnsplit[2]);
 
-					MoveFile(oldname.c_str(),newfilename.c_str());
+					if (FileOrDirExists (newfilename.c_str()))
+					{
+						if (MessageBox(g_hwndParent, __LOCALIZE("Filename already exists or file could not be offlined.\nIf you continue it might create a wrong source reference in your REAPER project!\nContinue?", "sws_DLG_165"), __LOCALIZE("Xenakios - Warning", "sws_DLG_165"), MB_OKCANCEL) == IDCANCEL)
```
cfillion, line 170: Offline status would not affect `FileOrDirExists`. Shouldn't this message just mention that the file already exists and asks whether to overwrite it?

Xenakios/TakeRenaming.cpp
```diff
@@ -223,40 +165,39 @@ void DoRenameTakeAndSourceFileDialog(COMMAND_T* ct)
...
+						{
+							Main_OnCommand(40101, 0); // online all media
+							continue;
+						}					
+					}
 					AddToRenameLog(oldname,newfilename);
```
cfillion, line 176: The `MoveFile` call is missing now.

nofishonfriday:
updated (I've amended the latest commit), thanks.

cfillion:
I can confirm it works as expected on macOS and Linux now.

I'm noticing playback cuts off while the message box is shown due to the offlining. Could this only be done when necessary, after the prompt?

Also renaming to the original filename (eg. clicking on OK instead of Cancel) shows overwrite the prompt + does an unnecessary peak rebuild. By the way is there a reason for not renaming the old peak file (wouldn't that avoid a potentially slow rebuild)?

nofishonfriday:
Good ideas, thanks.
Updated (I've made a new commit so you can see the changes at a glance, can then be squashed with previous before merging.)

Xenakios/TakeRenaming.cpp
```diff
@@ -223,40 +165,43 @@ void DoRenameTakeAndSourceFileDialog(COMMAND_T* ct)
...
+						if (!thesrcToRename || strcmp(SWS_GetSourceFileName(thesrcToRename), oldname.c_str()))
+							continue;
+
+						thesrcToRename->SetFileName(newfilename.c_str());
+						// rename peak file, could be stored outside audiofiles folder (user pref)
+						char oldPeakFn[SNM_MAX_PATH] = ""; char newPeakFn[SNM_MAX_PATH] = "";
+						GetPeakFileName(oldname.c_str(),     oldPeakFn, sizeof(oldPeakFn));
+						GetPeakFileName(newfilename.c_str(), newPeakFn, sizeof(newPeakFn));
+						if(!MoveFile(oldPeakFn, newPeakFn))
+							SNM_DeletePeakFile(oldname.c_str(), true); // fallback if rename didn't work for some reason
```
cfillion, line 191: I found three situations where this breaks:

- The target directory (first two letters of the hashed filename) may not exist yet when storing peaks in the alternate path.
- `newPeakFn` is always in the alternate path even when the setting is disabled. The peak file is then moved to the wrong location (unless the above prevents it). This can be fixed by using `GetPeakFileNameEx` with forWrite=true.
- The move fails if across different filesystems (at least on non-Windows). This must be done via copy + delete.

```cpp
void SNM_MovePeakFile(const char *oldMediaFn, const char *newMediaFn)
{
	char oldPeakFn[SNM_MAX_PATH] {}, newPeakFn[SNM_MAX_PATH] {};
	GetPeakFileName(oldMediaFn, oldPeakFn, sizeof(oldPeakFn));
	GetPeakFileNameEx(newMediaFn, newPeakFn, sizeof(newPeakFn), true);

	if (const char dirSep { WDL_remove_filepart(newPeakFn) }) {
		CreateDirectory(newPeakFn, nullptr);
		newPeakFn[strlen(newPeakFn)] = dirSep;
	}

	if (!MoveFile(oldPeakFn, newPeakFn)) {
		SNM_CopyFile(newPeakFn, oldPeakFn);
		SNM_DeleteFile(oldPeakFn, true); // move to trash on Windows
	}
}
```

```suggestion
						SNM_MovePeakFile(oldname.c_str(), newfilename.c_str());
```

Xenakios/TakeRenaming.cpp
```diff
@@ -114,103 +116,43 @@ void AddToRenameLog(string &oldfilename,string &newfilename)
...
 	if (thetakes.size()==0) return;
-	g_renameparams.takesToRename=(int)thetakes.size();
+	g_renameparams.takesToRename=thetakes.size();
 	g_renameparams.mode=2;
-	for (int i=0;i<(int)thetakes.size();i++)
+	for (size_t i=0;i<thetakes.size();i++)
 	{
 		PCM_source *thesrc=(PCM_source*)GetSetMediaItemTakeInfo(thetakes[i],"P_SOURCE",0);
-		if (thesrc && strcmp(thesrc->GetType(),"SECTION")!=0)
+		if (thesrc)
```
cfillion, line 132: This indentation level can be avoided:

```cpp
// if only renaming source
// skip takes which don't have a 'real' source
if (!thesrc || (ct->user == 0 && !SWS_GetSourceFileName(thesrc))
  	continue;
```

nofishonfriday:
updated, thanks for your patience.
I've also cleaned up the commit/msgs a bit.

Xenakios/TakeRenaming.cpp
```diff
@@ -114,149 +115,84 @@ void AddToRenameLog(string &oldfilename,string &newfilename)
...
 	vector<MediaItem_Take*> alltakes;
 	XenGetProjectTakes(alltakes,false,false);
 	XenGetProjectTakes(thetakes,true,true);
 	if (thetakes.size()==0) return;
-	g_renameparams.takesToRename=(int)thetakes.size();
-	g_renameparams.mode=1;
-	bool bChanges = false;
-	for (int i=0;i<(int)thetakes.size();i++)
+	g_renameparams.takesToRename=thetakes.size();
+	g_renameparams.mode=2;
```
cfillion, line 127: `mode` should be either 1 or 2 depending on `ct->user` to give the correct title to the dialog. It's always "Rename take and source file" now.

Xenakios/TakeRenaming.cpp
```diff
@@ -114,149 +115,84 @@ void AddToRenameLog(string &oldfilename,string &newfilename)
...
-		{
-			g_renameparams.curTakeInx=i+1;
-			string oldname;
-			oldname.assign(thesrc->GetFileName());
-			vector<string> fnsplit;
-			SplitFileNameComponents(oldname,fnsplit);
-			g_renameparams.OldName.assign(fnsplit[1]);
+		// if only renaming source
+		// skip takes which don't have a 'real' source
+		if (!thesrc || (ct->user == 0 && !SWS_GetSourceFileName(thesrc)))
```
cfillion, line 133: "Xenakios/SWS: Rename take source files... (no undo)" used to do nothing on MIDI items. Such sources have an empty filename, which is not skipped anymore.

```suggestion
		if (!thesrc || (ct->user == 0 && (!SWS_GetSourceFileName(thesrc) || !SWS_GetSourceFileName(thesrc)[0])))
```

nofishonfriday:
updated, thanks (tough one :D)

> EDIT: Looks like Tim will release next SWS tomorrow. I think this PR will need a bit of pre-release time.

Hm..wouldn't it make sense to add it in its current state? (= most likely fixed but untested vs. unfixed)

cfillion:
I'm worried it might have an uncaught regression (perhaps `SetFilename` behaves differently than `PCM_Source_CreateFromFile` in some edge cases?). I feel that would be worse in a rare featured version than fixing the old reverse/section bug, adding the overwrite warning feature and optimizing peak rebuilds.

Well, maybe I'm just too cautious...

nofishonfriday:
Ok, then how about adding a stripped down, _overwrite warning feature_ only version?
That's the most important part of this PR imo.

cfillion:
That would be perfect! I created a 2.12.1.3 branch earlier, you can add the overwrite warning straight to it.

nofishonfriday:
done. (first time I've pushed to SWS directly btw. :cold_sweat:)

cfillion:
I added another commit to wait until after the prompt to offline all media (otherwise it stays offline when cancelling) and avoid prompting when the new filename is identical.

If that's all right I'll squash the commits and update whatsnew later.

nofishonfriday:
Ah yes, I missed to take these changes over, thanks.

nofishonfriday:
updated to make the warning prompt default to Cancel rather than Ok (https://github.com/reaper-oss/sws/issues/1498#issuecomment-779848626)

PR #1493, opened 2021-01-07
GetArrangeWnd(), GetRulerWndAlt(), GetNotesView(), GetPianoView(): use windows control IDs instead of FindWindowEx()/SearchCildren()

nofishonfriday:
As discussed on Slack recently.

It also registers a temporary basic test action `SWS/NF: (temp) Test Window control IDs` to quickly check if the returned pointers are the same for both methods (will be removed and cleaned up before merging).

### DIFF
--- a/Breeder/BR_Util.cpp
+++ b/Breeder/BR_Util.cpp
@@ -2570,74 +2570,30 @@ HWND FindFloatingToolbarWndByName (const char* toolbarName)
 		}
 }
 
+// IDs of child windows are the same on all platforms
+// https://forum.cockos.com/showpost.php?p=2388026&postcount=5
+enum WndControlIDs               // caption:
+{
+	main_arrange   = 0x000003E8, // trackview
+	main_ruler     = 0x000003ED, // timeline
+
+	midi_notesView = 0x000003E9, // midiview
+	midi_pianoView = 0x000003EB  // midipianoview
+};
+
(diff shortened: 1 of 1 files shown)


cfillion:
Would you like to have this included in the v2.13 pre-release (minus the old code & test action)?

nofishonfriday:
If it ain't broke don't fix it. :D
Seriously, not that important, was first of all a proof of concept.
Questionable anyway if it'd bring any performance improvement I'd think, that would be the only reason to merge it I think.
What's your thought about it?

cfillion:
`GetNotesView_new` and `GetPianoView_new` are a few microseconds faster (50-80%) on my Windows VM. That won't have an impact, but I do like how simple your implementation is. 😄

nofishonfriday:
> but I do like how simple your implementation is.

Thanks. :)
So you think it's merge-worthy just for the code simplification?
Then I'd clean it up and undraft it.

nofishonfriday:
I cleaned up and undrafted it so a decision could be made if it should be merged or not for next release (plan to do that for my current other draft PRs too...).

cfillion:
> I'm not sure why Breeder first sets the HWND's to null and then checks for null afterwards

It's for running the lookup only the first time the function is called. On subsequent calls it just returns the cached value.

(Technically the `= NULL` is redundant, as variables with static storage duration are initialized by default.)

PR #1514, opened 2021-02-20
SWS/S&M: {Toggle arming of, Arm, Disarm} all active envelopes for selected tracks: also apply on send envelopes

### LINKED ISSUE #1513: after "hiding all envelopes for tracks" if we then "show all envelopes for tracks" and "Arm all active envelopes for selected tracks" volume sends will not get armed
Steps to reproduce:
- create 2 tracks
- route track 1 to track 2 to in order to create a send
- on track 1 set track automation mode to touch or latch or write 
- move the send knob while playing in order to create automation envelope
-  "toggle show active envelopes for tracks " in order to hide the envelope
-  "toggle show active envelopes for tracks" again in order for reaper to show the envelope again 
-  "Arm all active envelopes for selected tracks"

notice that this last action won't arm the sends envelope, it will only arm the volume, pan, plug ins automation etc... anything but the sends 

https://user-images.githubusercontent.com/79307447/108486908-8b243980-7296-11eb-8786-b68517e25fe7.mp4

nofishonfriday:
fixes #1513

+ReaScript: add NF_ArmEnvelope()
(I know we can do this via `BR_EnvSetProperties()` but it comes 'for free' with the added `SNM_ArmEnvParserPatcher_Env` and can be done directly on a `TrackEnvelope*`, no need to alloc `BR_Envelope*` first.)

There's currently a new bug introduced with my code with the `SWS/S&M: Toggle arming of all active envelopes for selected tracks` action in that when having a send track **and** the corresponding receive track selected when running this action, it toggles twice (on the send and on the receiving track) so effectively does nothing damnit.
Need to think about it (suggestions welcome :D).

update1:
Thinking about it, that these actions patch receive env's is not quite right imo (sure Envelope properties panel lists Receive envelopes, but visually they live on the sending tracks. Opinions?). Thus I made it now so that the actions only work on send envelopes (apart from the dedicated "Toggle arming of all receive ... envelopes for selected tracks" actions of course) which also solves the double toggling issue mentioned above.

update2:
Hm..I dunno, behaviour change is maybe not that good.
But the old behaviour together with this bugfix seems to lead to a handling problem.
If user has selected send and corressponding receive track and runs the "toggle arm active envs" action it's actually literally kinda 'correct' that it toggles twice (-> no-op) I'd say but of course not desired.
Think I'll stop the self-conversation for now and wait for comments/opinions. :)

### DIFF
--- a/ReaScript.cpp
+++ b/ReaScript.cpp
@@ -303,6 +303,7 @@ APIdef g_apidefs[] =
 	{ APIFUNC(NF_Win32_GetSystemMetrics), "int", "int", "nIndex", "Equivalent to win32 API GetSystemMetrics().", },
 	{ APIFUNC(NF_GetSWS_RMSoptions), "void", "double*,double*", "targetOut,windowSizeOut", "Get SWS analysis/normalize options. See <a href=\"#NF_SetSWS_RMSoptions\">NF_SetSWS_RMSoptions</a>.", },
 	{ APIFUNC(NF_SetSWS_RMSoptions), "bool", "double,double", "targetLevel,windowSize", "Set SWS analysis/normalize options (same as running action 'SWS: Set RMS analysis/normalize options'). targetLevel: target RMS normalize level (dB), windowSize: window size for peak RMS (sec.)", },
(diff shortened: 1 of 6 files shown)


nofishonfriday:
closing this as I'm not really active with SWS dev anymore (np if someone wants to take it as basis for a new PR)

PR #1531, opened 2021-04-03
Smart split actions: also handle razor edit areas (REAPER v6.24+)

### LINKED ISSUE #1528: SWS actions for "time selection" functions don't consistently work on razor edit areas
E.g. “Item: Split items at time selection” works on razor edit area.

But "SWS: Split items at time selection (if exists), play cursor (during playback), else at edit cursor" doesn't seem to recognize that a razor edit area in the same order of operations as "item selection," meaning the split only occurs on the edit cursor. (Same for “SWS: Split items at time selection (if exists), else at play cursor (during playback), or mouse cursor” and other similar SWS actions.)

nofishonfriday:
fixes #1528

+Add RazorEditArea.hpp/.cpp (wrapper class for razor edit area API)

### DIFF
--- a/Utility/RazorEditArea.cpp
+++ b/Utility/RazorEditArea.cpp
@@ -1,0 +1,125 @@
+/******************************************************************************
+/ RazorEditArea.cpp
+/
+/ Copyright (c) 2021 ReaTeam
+/
+/ Permission is hereby granted, free of charge, to any person obtaining a copy
+/ of this software and associated documentation files (the "Software"), to deal
+/ in the Software without restriction, including without limitation the rights to
+/ use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies
+/ of the Software, and to permit persons to whom the Software is furnished to
+/ do so, subject to the following conditions:
+/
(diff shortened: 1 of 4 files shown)


Utility/RazorEditArea.cpp
```diff
@@ -0,0 +1,125 @@
...
+	{
+		if (reas[i].m_envelopeGUID == "") // only check razor edit areas which aren't envelopes
+		{
+			MediaTrack* track = reas[i].m_track;
+			int itemCount = CountTrackMediaItems(track);
+			for (int j = 0; j < itemCount; j++)
+			{
+				MediaItem* item = GetTrackMediaItem(track, j);
+				int itemPos = GetMediaItemInfo_Value(item, "D_POSITION");
+				int itemEnd = itemPos + GetMediaItemInfo_Value(item, "D_LENGTH");
```
cfillion, line 92: Shouldn't these be doubles?

nofishonfriday: Updated, thanks.

PR #1538, opened 2021-05-12
ReaScript: BR_Win32_WritePrivateProfileString(): support deletion of a key by passing an empty string

nofishonfriday:


Breeder/BR_ReaScript.cpp
```diff
@@ -1096,6 +1096,8 @@ bool BR_Win32_WritePrivateProfileString(const char* sectionName, const char* key
 {
 	if (!strlen(keyName))
 		return 0;
+	if (value && !strcmp(value, "__delete"))
+		return !!WritePrivateProfileString(sectionName, keyName, nullptr, filePath);
```
cfillion, line 1100: Why not delete when the input value is null or empty? That would be closer to the Win32 equivalent (or is there a use case for storing empty values?).

Also, this line can be simplified as just `value = nullptr`.

PR #1555, opened 2021-09-09
rename actions "SWS: Goto/select {next,previous} marker/region" to "SWS: Go to/time-select {next,previous} marker/region"

nofishonfriday:
rename actions: Goto -> Go to, select region -> time-select region  (to not confuse with 'select region for rendering' which was added to REAPER later)

- rename "SWS: Goto/select {next,previous} marker/region" to "SWS: Go to/time-select {next,previous} marker/region"
report: https://forum.cockos.com/showpost.php?p=2479136&postcount=2956

- rename "SWS: Select {next,previous} region" to "SWS: Time-select {next,previous} region"

- rename "SWS/S&M: Go to/select region %d (obeys smooth seek)" to "SWS/S&M: Go to/time-select region %d (obeys smooth seek)"

#prettify MarkerList.cpp's g_commandTable (use spaces rather than tabs)


nofishonfriday:
While at it, maybe these should be renamed to "Go to/**time-**select...",
because selecting region has become a different meaning meanwhile:
https://forum.cockos.com/showthread.php?t=235259

cfillion:
Makes sense to me. There are also more "select region" actions at SnM.cpp:747 and MarkerList.cpp:690-691.

nofishonfriday:
I was trying to fix the indention but I don't get this tab indention thing. :/
Here it looks ok, but on GitHub it's screwed up:
https://github.com/reaper-oss/sws/pull/1555/files#diff-1b218784a31e9326a04c19f5f8ce29e4a16e658a0c9027206e0de57bab464770R690

How about we using spaces rather then tabs? :P

cfillion:
In MarkerList.cpp's g_commandTable? Yeah, it looks horrible. +1 for spaces!

nofishonfriday:
Actually I was wondering if we could maybe do away with the "We prefer: Hard tabs, equal to 4 spaces" coding guideline in the Wiki (or even turn it around for preferring spaces).
What do you think?

cfillion:
Didn't know about that hard tabs guideline. cfillion.cpp is violating it...  :scream:  I've abided by the "match the style of surrounding code" unspoken rule to far. (For my own code I'm a fan of 2-space indentation and spaces for alignment.)

The coding style is widely inconsistent throughout. There are tools (eg. clang-tidy) to enforce a specific style (there are a few widely-used ones), but considering how messy the code is I'm not sure it's worth rewriting all files at this point (and making `git blame` useless)...

nofishonfriday:
So - would you be ok with removing that rule?
I think the issue is that if we're using hard tab = 4 spaces but GitHub's default tab size = 8 spaces (afaik), so things can look messy (when e.g. used for alignment).

cfillion:
Yes. It's already not respected in many files (or parts of files!). A better guideline would be to match nearby code for indentation and general style. For alignment, spaces are objectively better as they aren't subject to the editor's settings.

nofishonfriday:
Just edited the Coding guidelines wikipage accordingly. Feel free to add/change. :)

Issue #1551, opened 2021-08-10
Bug report: "Play from mouse cursor position (perform until shortcut released)" breaks arrange view


maxhenderson43:
Noticed that calling the action "SWS/BR: Play from mouse cursor position (perform until shortcut released)" when the mouse cursor is outside the arrange viewport seems to completely break the arrange view. The grid, markers and regions get collapsed to a single point on the left-hand side and the items aren't visible. The edit cursor/selection position is shown as 0.nan / 0:00.00. 

<img width="1440" alt="Screenshot 2021-08-11 at 00 00 20" src="https://user-images.githubusercontent.com/88749991/128945679-dfbf034c-f6d8-4372-9a15-7a110bbf9b17.png">

Saving and reopening the project doesn't seem to fix this problem and I haven't found a way to restore the original state once this bug happens.

I haven't found this bug with the action "SWS/BR: Play from mouse cursor position" but it does happen with the other "perform until shortcut released" actions.

Using SWS 2.12.1.3, REAPER 6.33 and 6.34, Mac OS 11.2 and an M1 Macbook Pro 2020.

Would much appreciate any help on this issue :)

jeremybernstein:
Can you unpack "when the mouse cursor is outside the arrange viewport"? I can't reproduce under macOS 10.15 (Intel 64-bit), but I can imagine that I'm also not understanding how to reproduce. How are you triggering the action?

nofishonfriday:
I can reproduce on Windows.
Steps:
- open action list (Actions > Action list...)
- assign a keyboard shortcut to said action
- resize Reaper so that it doesn't cover full screen (but make sure it's still focused so it receives keyboard shortcuts)
- move mouse cursor outside of the Reaper window
- trigger the action via the assigned keyboard shortcut

![broken arrange](https://i.imgur.com/dvHCspj.jpg)

jeremybernstein:
I've tried this on a fresh REAPER 6.3.6 install + SWS 2.12.1.3 on a Mac M1 and Mac Intel machine, and cannot reproduce. The window definitely has keyboard focus (confirmed that the same key works when playing from edit position). 

There must be some step I'm missing. Maybe because I'm using an empty project to start (with a few tracks and empty MIDI items), or something short? Stumped.

jeremybernstein:
Nor can I reproduce on a clean install of REAPER on Windows 10 (albeit hosted in a VM -- I have a real PC in my studio, and I can try to reproduce on that system, as well).

[...]

PR #1559, opened 2021-09-24
jb/1551

jeremybernstein:
This changeset addresses the issue raised in #1551 ("Play from mouse cursor position (perform until shortcut released)" breaks arrange view). The cleanup code path (!init) would be run, even if init failed, causing a specific problem in the event that "Scroll view to edit cursor on stop" is true, but potentially causing other problems.

A second commit ensures that the static variables are correctly initialised at the top of initialisation and fixes a potential memory leak.

### DIFF
--- a/Breeder/BR_Misc.cpp
+++ b/Breeder/BR_Misc.cpp
@@ -208,7 +208,14 @@ static bool MousePlaybackInit (COMMAND_T* ct, bool init)
 
 			s_itemMuteState = new vector<pair<GUID,int> >;
 			if (!s_itemMuteState)
+			{
+				if (s_trackSoloMuteState)
+				{
+					delete s_trackSoloMuteState;
+					s_trackSoloMuteState = NULL;
+				}
 				return false;
+			}
 
 			int count = CountTrackMediaItems(trackToSolo);
 			for (int i = 0; i < count; ++i)


jeremybernstein:
This failing check appears to be a configuration problem on AppVeyor: the Linux cc can't compile a test program, and appears to be unrelated to my changeset.

jeremybernstein:
How do I rerun the check against the 2.13 branch?

nofishonfriday:
An empty (without any changes) `git push --force` should trigger a check rerun I think.

cfillion:
`MousePlaybackInit` should return false when no branches are taken.

```cpp
Breeder/BR_Misc.cpp: In function ‘bool MousePlaybackInit(COMMAND_T*, bool)’:
Breeder/BR_Misc.cpp:377:1: warning: control reaches end of non-void function [-Wreturn-type]
```

cfillion:
Merging into the work-in-progress v2.13 branch closed the PR early because it was based on it... (The bugfix seemed ready to me, but if you'd like to make further changes, feel free to make a new PR.)

I'm thinking of skipping the memory leak commit, because (unless I'm missing something):

- The improved `else` block made it redundant as `MousePlaybackInit` is called again with `init=false` after initialization fails to perform the cleanup.
- That `delete` can't be reached as `s_itemMuteState` would only be null if the memory allocation above failed, which would raise an exception and crash. (Maybe the original code's intended to use the non-throwing overload of `new`?)

jeremybernstein:
@cfillion that makes sense to me -- now that the cleanup block is being called in all cases and the "leaking" pointers deleted, it's not necessary. As long as the calling code doesn't change (and `MousePlaybackInit` continues to be called twice), it is redundant.

PR #1562, opened 2021-10-01
Add ability for region playlist crop and append to include tempo mark…

### LINKED ISSUE #1462: Major issue with SWS playlist and tempo changes
I just discovered the Playlist in SWS and I love it. I can rearrange regions etc and make structures quickly... only problem I am having is: When I make an arrangement I like in playlist and then I click the EDIT PROJECT button and I create a new arrange (either in new tab or overwrite the current one) the regions get reordered as from the playlist BUT all the tempo changes aren't moved with the region. Actually all the tempo changes get deleted and there is a common tempo for all region which isn't correct! Am I doing something wrong here?

thank you 

PS 
another issue: if I have the tracks grouped the region does't cut them within the border but copy the full lenght of the audio file creating overlaps and a general mess. It took me a while to figure out I had to ungroup tracks to make it work. Still tempo changes aren't copied though

Sonicville360:
…ers fixes #1462

Pretty new to this so feel free to review and comment anything that could be done better.
As commit name says, this makes it that the region playlist "Append" and "Crop" commands also take in account tempo markers.
Tested on windows on a quite complex session successfully.
The whole part with chunks is because we needed to have also the "Metronome Pattern" data duplicated, but it isn't accessible via standard getters and setters.

### DIFF
--- a/SnM/SnM_Marker.cpp
+++ b/SnM/SnM_Marker.cpp
@@ -384,7 +384,7 @@ bool GetAllTempoMarkers(WDL_PtrList<void>* _tempoMarkers, ReaProject* _proj)
 		while (token != NULL)
 		{
 			RawTempoMarkerData newTempoMarker = RawTempoMarkerData();
-			if (sscanf(token, "PT %lf %lf %d %u %d %d %d %s %u %u", &newTempoMarker.position, &newTempoMarker.bpm, &newTempoMarker.linearBool, &newTempoMarker.beatDivision, &newTempoMarker.intData4, &newTempoMarker.settingsBitmask, &newTempoMarker.intData6, newTempoMarker.stringData7, &newTempoMarker.uintData8, &newTempoMarker.metronomePattern) >= 2)
+			if (sscanf(token, "PT %lf %lf %d %u %d %d %d %s %u %u %d", &newTempoMarker.position, &newTempoMarker.bpm, &newTempoMarker.linearBool, &newTempoMarker.beatDivision, &newTempoMarker.selectionBool, &newTempoMarker.settingsBitmask, &newTempoMarker.bezierTension, newTempoMarker.quantizationSettings, &newTempoMarker.metronomePatternL32, &newTempoMarker.metronomePatternH32, &newTempoMarker.beatBase) >= 2)
 			{
 				tempoDataMap[index] = newTempoMarker;
 				index++;
@@ -433,7 +433,7 @@ bool GetTempoMarkersInInterval(WDL_PtrList<void>* _tempoMarkers, ReaProject* _pr
 		while (token != NULL)
 		{
 			RawTempoMarkerData newTempoMarker = RawTempoMarkerData();
-			if (sscanf(token, "PT %lf %lf %d %u %d %d %d %s %u %u", &newTempoMarker.position, &newTempoMarker.bpm, &newTempoMarker.linearBool, &newTempoMarker.beatDivision, &newTempoMarker.intData4, &newTempoMarker.settingsBitmask, &newTempoMarker.intData6, newTempoMarker.stringData7, &newTempoMarker.uintData8, &newTempoMarker.metronomePattern) >= 2)
+			if (sscanf(token, "PT %lf %lf %d %u %d %d %d %s %u %u %d", &newTempoMarker.position, &newTempoMarker.bpm, &newTempoMarker.linearBool, &newTempoMarker.beatDivision, &newTempoMarker.selectionBool, &newTempoMarker.settingsBitmask, &newTempoMarker.bezierTension, newTempoMarker.quantizationSettings, &newTempoMarker.metronomePatternL32, &newTempoMarker.metronomePatternH32, &newTempoMarker.beatBase) >= 2)
 			{
 				tempoDataMap[index] = newTempoMarker;
 				index++;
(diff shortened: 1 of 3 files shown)


nofishonfriday:
Thanks for the pull request, nice to see some new contributers. :thumbsup:

Sonicville360:
Mmmh, anyone has an idea why it specifically doesn't compile on the Linux x86 32-bit test? I'm not sure I understand the error log.

nofishonfriday:
> 
> 
> Mmmh, anyone has an idea why it specifically doesn't compile on the Linux x86 32-bit test? I'm not sure I understand the error log.

Don't worry about it, it's an auto-build misconfiguration issue, already fixed in the v2.13 branch (https://github.com/reaper-oss/sws/commit/b6e80c0797281425cb1e395f6f0f6f2f9705c9ca).

nofishonfriday:
@Sonicville360 Sorry for the Iooong delay. We're currently in the process of [preparing](https://github.com/reaper-oss/sws/pull/1683) a new SWS pre-release. I looked briefly at this PR, ts_beatbase (11th `PT` value) is missing in the `RawTempoMarkerData`.
It shows up when setting BPM basis to something other than 1/4 in the Edit Tempo/Time signature marker dialog in current REAPER version.
Also the meanings of the other values could added in your [comments](https://github.com/reaper-oss/sws/pull/1562/files#diff-d606f62593e442e291b7ca675f3874a646f9e29cb760fc2c9c7f027ca743f065R75) for documentation purpose [as mentioned by Justin](https://www.askjf.com/index.php?q=6429s). 
Would you be up for tweaking the PR?

SnM/SnM_Marker.cpp
```diff
@@ -345,3 +345,228 @@ void InsertMarker(COMMAND_T* _ct)
...
+			RawTempoMarkerData newTempoMarker = RawTempoMarkerData();
+			if (sscanf(token, "PT %lf %lf %d %u %d %d %d %s %u %u %d", &newTempoMarker.position, &newTempoMarker.bpm, &newTempoMarker.linearBool, &newTempoMarker.beatDivision, &newTempoMarker.selectionBool, &newTempoMarker.settingsBitmask, &newTempoMarker.bezierTension, newTempoMarker.quantizationSettings, &newTempoMarker.metronomePatternL32, &newTempoMarker.metronomePatternH32, &newTempoMarker.beatBase) >= 2)
+			{
+				tempoDataMap[index] = newTempoMarker;
+				index++;
+			}
+			token = strtok(NULL, "\n");
+		}
+
+		for (int i = 0; i <= CountTempoTimeSigMarkers(_proj); i++)
```
cfillion, line 395: Off-by-one error: `GetTempoTimeSigMarker(count)` does not exist.

cfillion (commented):
Code quality needs work. It's very verbose/C-like, uses the wrong data containers, checks for impossible scenarios (null or non-empty output list) while having memory leaks and buffer overflows.

Can probably be optimized by 1) removing duplicate steps like partially editing tempo markers using the API before overwriting the whole envelope chunk 2) only doing it once per operation instead of once per region 3) iterating/searching through lists fewer times.

It doesn't insert a marker at the beginning of the copied region if the tempo at the destination location is different than at the source.

TempoMarker and RawTempoMarkerData should probably be merged. Most of the data overlap and doesn't need to be separate.

(Nitpick: The `std` namespace is inconsistently specified or omitted.)

(I remember seeing an older version of this PR create corrupted markers in REAPER v5 when I first tested a while ago. Can't duplicate with today's version though. Maybe it's thanks to the change in default value from -1 to 0?)

SnM/SnM_Marker.cpp
```diff
@@ -345,3 +345,228 @@ void InsertMarker(COMMAND_T* _ct)
...
+		double dBeatPosOut;
+		double dBpmOut;
+		int timesig_numOut;
+		int timesig_denomOut;
+		bool bLinearTempoOut;
+
+		map<int, RawTempoMarkerData> tempoDataMap;
+
+		MediaTrack* masterTrack = GetMasterTrack(_proj);
+		TrackEnvelope* tempoEnvEx = GetTrackEnvelope(masterTrack, 0);
```
cfillion, line 377: This may not give the tempo envelope if there are other envelopes on the master track.

```
GetTrackEnvelopeByName(masterTrack, "Tempo map")
```

cfillion:
I wrote an alternative implementation of this PR in #1694. It uses only chunk manipulation.

> On a side note, this push still has a small "bug" which I don't know how to solve. When appending or cropping the project, the metronome patterns are duplicated correctly but they aren't "applied". I have to save the new project, close and reopen it to hear the correct metronome pattern. I guess I have to reload something after the marker duplication but I wouldn't know what.

Confirmed. I think this is a REAPER bug: it can be duplicated with the script below.

```lua
local chunk = [[
<TEMPOENVEX
EGUID {EA3AB88C-668F-2147-90D8-591728D4085D}
ACT 1 -1
VIS 1 0 1
LANEHEIGHT 0 0
ARM 0
DEFSHAPE 1 -1 -1
PT 0.000000000000 120.0000000000 1
PT 2.000000000000 120.0000000000 1 0 0 8 0 "" 2863311530 2863311497
>
]]

local mtr = reaper.GetMasterTrack(nil)
local env = reaper.GetTrackEnvelopeByName(mtr, "Tempo map")
reaper.SetEnvelopeStateChunk(env, chunk, false)
```

Sonicville360:
Hey @cfillion! Thanks for the thorough review.
As you have already written an alternative implementation in #1694 should I spend time correcting this or should I just delete the PR and branch?

cfillion:
I found a workaround for the metronome pattern issue: calling `SetTempoTimeSigMarker` on any marker after setting the envelope chunk applies the new patterns.

> As you have already written an alternative implementation in https://github.com/reaper-oss/sws/pull/1694 should I spend time correcting this or should I just delete the PR and branch?

Please keep the branch intact, in case my version doesn't handle some edge cases, it may be useful to refer to this one as reference.

(Pushed some fixes to my branch related to item timebase, since setting points via the chunk doesn't update everything on REAPER's side, I wouldn't be surprised if more special cases are found. More testing required...)

PR #1563, opened 2021-10-01
Fix persisting colors on macOS [Fix #1564]

### LINKED ISSUE #1564: Custom colors are not persisted after reaper restart on macOS
When you create custom colors and then restart Reaper, the colors are not saved. 
You'll end up with the default black colors: 

image

System info:
macOS Big Sur 11.5.1
Macbook Pro M1
Reaper v6.36
SWS Extensions v2.12.1.3

This has been an issue since at least 2018. See https://forums.cockos.com/showthread.php?t=205542

**Steps to repro**
1. Open Reaper
2. Open Color Management
3. Load custom file (unzip attached file and load it)
4. Right-click on a track and take note of the available track colors
5. Restart Reaper
6. Right-click on a track

**Expected:** The previous colors are shown
**Actual:** The default black colors are shown

karlingen:
We are not persisting the custom colors to REAPER.ini on macOS for some reason. The user has to restore the custom colors every time Reaper is restarted. This fixes #1564 

It would be nice with an explanation as to why we weren't doing this in the first place and if I might have missed something vital.

### DIFF
--- a/Color/Color.cpp
+++ b/Color/Color.cpp
@@ -45,9 +45,8 @@ static bool g_bRecRedRuler = false;
 
 void UpdateCustomColors()
 {
-#ifndef __APPLE__
 	GetPrivateProfileStruct("REAPER", "custcolors", g_custColors, sizeof(g_custColors), get_ini_file());
-#else
+#ifdef __APPLE__
 	GetCustomColors(g_custColors);
 #endif
 }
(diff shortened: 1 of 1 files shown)


cfillion:
On macOS the custom colors should be saved into the user's global favorite color palette rather than the in the .ini. The root cause of the bug is `SetCustomColors` failing to save the colors to the system palette on high sierra and newer.

I wrote a potential fix, could you give 60f27720b2ad94942957e3b1d0162d542830bf2d a try?

karlingen:
@cfillion Just gave it a go. It works! It also fixed the custom color picker 😃 
Awesome work man!

Can we get a PR up for it? I'll close this one

karlingen:
@cfillion Also for future reference, where are all of the `NSFavoriteColorsStore` and related APIs documented?

cfillion:
> Where are all of the NSFavoriteColorsStore and related APIs documented?

They aren't documented as far as I could find. Seems to be a private Apple API (so it could break anytime). I found about those functions and classes by reverse-engineering how the system color picker saves the colors.

PR #1567, opened 2021-10-06
Additional track gestalts for autocolor

jeremybernstein:
Adding some useful track identification parameters for use in the Autocolor dialog. I've been using these for a while and they're reliable. Question is simply whether they are useful for others beside myself.

### DIFF
--- a/sws_extension.cpp
+++ b/sws_extension.cpp
@@ -485,9 +485,8 @@ public:
 	const char *GetDescString() { return ""; }
 	const char *GetConfigString() { return ""; }
 
-	bool m_bChanged;
 	int m_iACUpdateFlags;
-	SWSTimeSlice() : m_bChanged(false), m_iACUpdateFlags(ACUpdate_None) {}
+	SWSTimeSlice() : m_iACUpdateFlags(ACUpdate_None) {}
 
 	void Run() // BR: Removed some stuff from here and made it use plugin_register("timer"/"-timer") - it's the same thing as this but it enables us to remove unused stuff completely
 	{          // I guess we could do the rest too (and add user options to enable where needed)...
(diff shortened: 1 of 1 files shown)


cfillion:
It works nicely here on both v5 and v6 now. https://i.imgur.com/TPqI444.gif

Using spaces instead of the hyphen would blend better with the existing filters. Also, I think the audio and MIDI I/O filters should be together, with instrument either above or below.

Diff

jeremybernstein:
Nice, I have no objection to either. I think I'd put instrument above, just looks nicer. The stored setting appears to use the strings, so any changes in that regard (hyphen -> space) will require a reselection to make it work again.

Color/Autocolor.cpp
```diff
@@ -812,6 +812,41 @@ void ApplyColorRuleToTrack(SWS_RuleItem* rule, bool bDoColors, bool bDoIcons, bo
...
+					else if (strcmp(rule->m_str_filter.Get(), cFilterTypes[AC_AUDIOIN]) == 0)
+					{
+						int input = *(int*)GetSetMediaTrackInfo(tr, "I_RECINPUT", NULL);
+						if (input >= 0 && !(input & 4096)) { // !none && !MIDI
+							bMatch = true;
+						}
+					}
+					else if (strcmp(rule->m_str_filter.Get(), cFilterTypes[AC_AUDIOOUT]) == 0)
+					{
+            int hwouts = GetTrackNumSends(tr, 1);
```
cfillion, line 824: Indentation is messed up in this line (spaces vs tabs).

jeremybernstein:
Thanks, my bad. While I was fixing it, I fixed it in a couple of other places in the same file (where it wasn't my fault), as well. :-)

PR #1583, opened 2021-11-30
Add new ReaConsole command for writing custom extended state

jtackaberry:
This allows for generalized scripts that take "arguments" by creating a cycle action that first uses this new ReaConsole command to prime a value in REAPER's non-persisted state, and then calling a custom script that subsequently reads the value.

As discussed here: https://forum.cockos.com/showthread.php?p=2501831#post2501831


nofishonfriday:
Thanks for the PR (nice to see new SWS contributers :)).
Btw. we've ditched updating whatsnew.txt with PRs (see https://github.com/reaper-oss/sws/wiki/Format-of-the-whatsnew.txt).

jtackaberry:
> Thanks for the PR (nice to see new SWS contributers :)).

And now that I know how to scratch that particular itch, you might see more PRs from me.  Ironically, finding a working `reaper_plugin.h` was more work than this actual PR. :)

>  Btw. we've ditched updating whatsnew.txt with PRs (see https://github.com/reaper-oss/sws/wiki/Format-of-the-whatsnew.txt).

Good to know, thanks. Do you need me to revert the change in this PR?

nofishonfriday:
>  Ironically, finding a working reaper_plugin.h was more work than this actual PR. :)
From v2.13 it's fetched directly from Justin's GitHub repo (https://github.com/reaper-oss/sws/commit/7af31cbcc52a5903ca36bdf11a2b9e9e4bfa98dd), so that should ease things. :)

> Do you need me to revert the change in this PR?
Sorry, please see my edit.

jtackaberry:
> From v2.13 it's fetched directly from Justin's GitHub repo ([7af31cb](https://github.com/reaper-oss/sws/commit/7af31cbcc52a5903ca36bdf11a2b9e9e4bfa98dd)), so that should ease things. :)

Aha, indeed, that's a proper submodule now.  I think what happened is that I followed the [build instructions](https://github.com/reaper-oss/sws/wiki/Building-the-SWS-Extension) and ended up playing a bit of a game of whack-a-mole, not realizing things had changed in the v2.13 branch and I need to init git submodules until *after* I whacked the particular moles for `reaper_plugin{,_functions.h}` (and the missing #define for `PCM_SOURCE_EXT_GETBITRATE`).  And at that point I didn't clue in that vendor/reaper-sdk was a thing because it wasn't the next problem I needed to solve. :)

I don't have access to change the build wiki, but if step 2 were updated to note that for the v2.13 branch all you need to do in that step is `git submodule update --init` it might save the next developer a bit of head-scratching.  (Especially if SWS is clamoring for new contributors, anything to lower the barrier to entry is good.)  In any case the build process is much improved with the use of the reaper-sdk submodule.

nofishonfriday:
Sorry for the hassle.
As by experience it's (sadly) not very frequent that new SWS contributers show up, keeping the Wiki up to date sometimes falls a bit behind. :/

>  if step 2 were updated to note that for the v2.13 branch all you need to do in that step is git submodule update --init it might save the next developer a bit of head-scratching.
Done.

> Though I guess once v2.13 is promoted that won't be necessary, as git clone --recursive will take care of it. Step 2 can be removed entirely.
Good point, thanks for the reminder, will try to keep it in mind for after v2.13 becomes official.

jtackaberry:
> Done.

Thanks!

> I think it might still be better generating it from from current REAPER.

If we can't depend on Justin's repo to be current, then presumably we should be maintaining an up-to-date copy in this repo, if nothing else than for the CI tests?

PR #1301, opened 2020-01-30
New Project Set List

ruben-sandoval:
Hi,

I'm not a C++ programmer... most of the code is just a copy of Project Management, the main difference is that Set List will not auto-load all the songs.
I also copy code from Region Playlist to flag the "current song" playing and "next" song to play.

Code worked fine on Windows and OSX.

From whatsnew.txt
Project Set List:
+Project Set List will let users play multiple Reaper Projects in sequence.
+Shortcuts
 - [A] Add new song.
 - [D] Delete selected song(s) from the set list.
 - [N] Load next song in the set list.
 - [P] Move song(s) up in the set list.
 - [M] Move song(s) down in the set list.
 - [C] Clear set list.
 - [L] Load set list.
 - [S] Save set list.
 - [O] Auto load next song in the set list.
 - [Y] If "Auto load" was enabled, "Y" will auto play the next song.
 - [Enter] Load song. If song is playing and "Auto Load" is enabled, hitting [Enter] will flag the song as the next song to play.
 - Mouse Drag - Move song(s) in the list.
 - Windows user will be able to insert add songs from the open file dialog.
 - [Shift+Up / Dn] - Select multiple songs.

Thanks,
Ruben

PR #1589, opened 2021-12-16
Release v2.13

cfillion:
This merges all pull requests that were ready for a new pre-release. Not included are the drafts (#1449, #1467, #1493, #1514), incomplete (#1291, #1444) and unreviewed PRs (#1301, #1562).

If there's nothing to add/changelog typos/other changes to make, all that's left is amending the last commit with the correct release date in the whatsnew then merging the branch into master.

- The submodules WDL, TagLib are updated and reaper-sdk added (run `git submodule update --init`)
- reaper_plugin_functions.h is read from the new official SDK repository if it doesn't exist in the vendor subfolder (delete to migrate to the new SDK)
- The value of some cached CMake variables changed, recreating the build tree might be necessary if it was created against the master branch (IDEs should do this automatically)


nofishonfriday:
Yay, thanks for the work!

Nitpick on the whatsnew:
> +Add NF_ functions for getting, setting and clearing global/project/track selection actions

could add: (issue 974)

cfillion:
> could add: (issue 974)

Added.

Edgemeal:
This is weird, I set name of a new track so SWS v2.12 auto colors it, then script optionally gets that track color,  darkens the color and sets the track color, but with v2.13 instead of the track color getting darker it changes to a black color. If I force my script to wait a bit (using defer loop) then the track darkening works as it does in current SWS release. Very weird! 

**TESTED on: Win10_x64, REAPER 6.41 & v6.43_RC4
sws-2.13.0.0-Windows-x64-v2.13-8e57814d.exe**

example,
NewTrackName = "Drums" --  SWS auto colors track to a light blue.
ControlKeyIsDown = false -- if true darken track color, but it doesn't work as expected with SWS 2.13.

function DarkenColor(trk_colr) -- simple color darkener
  local b,g,r = reaper.ColorFromNative(trk_colr)
  if b - 50 < 0 then b = 0 else b = b - 50 end
  if g - 50 < 0 then g = 0 else g = g - 50 end
  if r - 50 < 0 then r = 0 else r = r - 50 end
  return reaper.ColorToNative(b,g,r) 
end

track = reaper.GetTrack(0,0) -- SWS auto colors the track when named.
reaper.GetSetMediaTrackInfo_String(track, "P_NAME", NewTrackName, true ) 

-- Optionally make the track a slightly darker
if ControlKeyIsDown then reaper.SetTrackColor(track, DarkenColor(reaper.GetTrackColor(track))) end
--  with SWS v2.13 ^ this makes the track black instead of a darker blue.

Edgemeal:
Guessing the problem is, if you try to get the color of a track right after SWS has auto colored it , 2.13 now returns the wrong value?

cfillion:
> If I force my script to wait a bit (using defer loop) then the track darkening works as it does in current SWS release.

Auto-color was indeed changed to be asynchronous (triggering on the next defer tick). This was done for performance reasons for the new track I/O modes and was applied to all modes for consistency.

I've replaced that with a different solution (downside is it's unclear if it cover all cases, relies on REAPER sending redundant control surface notifications in a particular order).

cfillion:
I've updated the release date to tomorrow.

nofishonfriday:
Just a (maybe silly) idea...how about following the Reaper release tradition and letting users announce new SWS releases in the forum (the last ones you announced iirc).
Not sure if it would happen though, maybe nobody notices or bothers to post lol.

cfillion:
I suppose the majority of users with pre-release notifications enabled aren't necessarily active on the forum anyway...

PR #1596, opened 2022-01-03
Don't unregister commands when REAPER quits

nofishonfriday:
according to Justin's comment:
"SWS should be updated to not unregister IDs it doesn't own (or not unregister IDs at all, really)"
https://forum.cockos.com/showthread.php?p=2509487#post2509487

Would probably supersede #1591.
I've left `SWSUnregisterCmd()` in because it's also used by Cycle actions (`FlushCycleactions()`).

### DIFF
--- a/sws_extension.cpp
+++ b/sws_extension.cpp
@@ -335,11 +335,6 @@ COMMAND_T* SWSUnregisterCmd(int id)
 	return NULL;
 }
 
-void UnregisterAllCmds() {
-	for (int i=g_iFirstCommand; i<=g_iLastCommand; i++)
-		SWSUnregisterCmd(i);
-}
-
 #ifdef ACTION_DEBUG
 void ActionsList(COMMAND_T*)
 {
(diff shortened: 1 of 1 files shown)


cfillion:
I think it's better to still properly unregister anything that was registered when unloading, even though omitting it is technically OK from REAPER's side.

I haven't checked, but wouldn't there be a chance of a script's `atexit` running after SWS has partially unloaded, running a SWS action leading to a use-after-free?

nofishonfriday:
> I think it's better to still properly unregister anything that was registered when unloading, even though omitting it is technically OK from REAPER's side.

Hm...to me it seems to contradict with what Justin is saying ("not unregister IDs at all, really").
Could you elaborate on your reasoning?

cfillion:
I meant that while it can be safe to omit unregistering extension actions from REAPER's side, it might not be so from SWS' side. (Plus in general it's good practice for software to cleanup after itself, even if technically the OS does it anyway.)

I thought for instance a buggy script doing

```lua
function loop() reaper.defer(loop) end
reaper.defer(loop)

reaper.atexit(function()
  reaper.Main_OnCommand(reaper.NamedCommandLookup('_SWSSNAPSHOT_OPEN'), 0)
end)
```

...would lead to using `g_pSSWnd` after it has been freed.

But I've just checked, that isn't actually the case, as action callbacks are disabled. So the above is an non-issue.

https://github.com/reaper-oss/sws/blob/b8bb9af7935ae2ef6e92f17faeff54064658e447/sws_extension.cpp#L658-L659

nofishonfriday:
Ok, closing this one. :)

PR #1604, opened 2022-01-25
Expose BR SimulateMouseClick() to ReaScript

nofishonfriday:
request: https://forum.cockos.com/showthread.php?p=2519895


amagalma:
Thank you nofish!! :)

It seems to work...  at least it sends the required lmousebutton messages.. but sometimes it seems that this is not enough. Have a look at the licecap:
https://i.ibb.co/pnwjLyp/show.gif

The [SWS/S&M: Left mouse click at cursor position (use w/o modifier) action](https://github.com/reaper-oss/sws/blob/4ca0494874c3b38cac5372bc56505835e46150ac/SnM/SnM.cpp#L518) seems to always work, so perhaps it is not just a matter of exposing the SimulateMouseClick function but of making [another version of this one](https://github.com/reaper-oss/sws/blob/4ca0494874c3b38cac5372bc56505835e46150ac/SnM/SnM_Misc.cpp#L437) that accepts coordinates.

nofishonfriday:
@amagalma
> but of making another version of this one that accepts coordinates.

This action is [Win only](https://github.com/reaper-oss/sws/blob/4ca0494874c3b38cac5372bc56505835e46150ac/SnM/SnM.cpp#L517) unfortunately and I don't think it's a good idea to expose a Win only function to ReaScript + I don't have the sufficient knowledge to translate it to Mac/Linux (specifically [this code](https://github.com/reaper-oss/sws/blob/4ca0494874c3b38cac5372bc56505835e46150ac/SnM/SnM_Misc.cpp#L413) which is called by the action to make it synchronous).

Out of my league, sorry, so I'm closing this PR.
All I can suggest you is open an 'official' FR issue here and hope someone else chimes in. :/

amagalma:
No problem! Thank you very much for your time and effort! :)

PR #1662, opened 2022-07-18
acendan: Squashed commit. Add ReaScript API: AC_GetMediaFileCoverImage()

acendan:
1. acendan: Initial commit. Add ReaScript API: AC_GetMediaFileCoverImage()
2. Removed <codecvt>, was previously relying on wstrings for file I/O, but no longer needed for character encoding conversions.
3. Swap std::copy for snprintf. Add newline to end of files.

Tested: Win10, OSX 10.14.6, Linux Mint

PR #1653, opened 2022-07-14
Actions: add SWS: Scroll {left,right} {75%,100%}

nofishonfriday:
Request: https://forum.cockos.com/showthread.php?t=268728

PR #1645, opened 2022-06-26
ReaScript: add NF_Base64_Decode(), NF_Base64_Encode()

nofishonfriday:
closes #778

Not sure I'm doing this buffer (re-)sizing stuff correctly but looks like it works. 
Also not sure about padding, seems like this version in SWS doesn't do it or am I doing s.th. wrong or should we use another version?
base64

PR #1606, opened 2022-02-04
use GitHub issue forms for opening issues

nofishonfriday:
We could use GitHub [issue forms](https://github.blog/changelog/2021-06-23-issues-forms-beta-for-public-repositories/) for opening (more structered) issues here.

[Live demo](https://github.com/nofishonfriday/sws/issues/new/choose)

About the actual content of the templates I'm open to discussion as this was ~~stolen~~ [inspired](https://github.com/TeamNewPipe/NewPipe/pull/7773)  from another repo anyway (thanks!). :)
E.g. I'm not sure if the
> IF YOU DON'T FILL IN THE TEMPLATE PROPERLY, YOUR ISSUE IS LIABLE TO BE CLOSED.

paragraph is maybe a bit too strict...

edit:
Maybe the automatic `bug` labeling should also be removed and still done manually by the maintainers after reproduction?

PR #1683, opened 2022-10-11
Versino 2.13.2

cfillion:
- [x] #1635
- [ ] #1637
- [x] #1639
- [x] #1642
- [ ] #1645
- [x] #1653
- [ ] #1662
- [x] #1665
- [x] #1677
- [x] #1678
- [x] #1682


nofishonfriday:
@cfillion 
I'm still a fan of #1606.
I think we didn't discuss this conclusively for last release, or did we?
If you say meh fine with me also, it just doesn't need to stay open forever (so either merge or close is fine).

nofishonfriday:
Tasks complete, thanks for the work. :)
(Now waiting for Tim...)

cfillion:
At this point I'm starting to think we should ship the 1.13.2 pre-release without waiting for the 1.13.1 stable to come out first. More so with the vzoom2 breakage coming in a forthcoming REAPER version.

That would mean giving up having the stable release appear in the master branch's whatsnew, commit history and in the pre-release download page. A "shadow" stable release could be done that way by just directly uploading builds from #1714 to the website.

nofishonfriday:
I can't really judge this so I'll leave it up to you. :)
I agree though it would be good to get the 2.13.2 pre out asap, personally I'm already using the current build here since yesterday so I can use SWS (drag) zoom vertically again with the latest REAPER pres. 

Btw. maybe could include #1721 too in the 2.13.2 pre if you've come up with a fix meanwhile, as it's an accessibility issue?

cfillion:
> Btw. maybe could include https://github.com/reaper-oss/sws/issues/1721 too in the 2.13.2 pre if you've come up with a fix meanwhile, as it's an accessibility issue?

Yes, found the solution, cleaning it up for a PR right now. 😄

PR #1760, opened 2023-05-22
Fix 'Xenakios/SWS: Choose new source file for selected takes...' not updating take name

### LINKED ISSUE #1758: Update item Name after running Xenakios/SWS: Choose new source file for selected takes...
after running this action the name of the item in the project doesn't change. kinda hard to track which have been replaced.

Other methods of changing source files, drag & drop, or project bay, do change out the name as well.

nofishonfriday:
closes #1758

### DIFF
--- a/Xenakios/ItemTakeCommands.cpp
+++ b/Xenakios/ItemTakeCommands.cpp
@@ -27,6 +27,7 @@
 
 #include "stdafx.h"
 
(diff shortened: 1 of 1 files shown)


cfillion:
I think it should check whether the old take name matches the old source's filename so that it preserves custom take names like 'Choose new file' in the take properties dialog.

Issue #1761, opened 2023-05-23
New Reaper track spacers cause issues with SWS mouse dependent features


musicbynumbers:
Not sure if this is something Reaper devs can fix but the new track spacers mess up the vertical positioning on a few SWS things when used in conjunction with the mouse position. 

Mainly things like: 

-contextual toolbars (showing menu under mouse) 
-previewing item under mouse
-SWS zoom tool (can no longer draw a box)

There's probably others and this might also affect reascripts etc too.

I've mentioned it here on the reaper pre release forum too.
https://forum.cockos.com/showpost.php?p=2679292&postcount=19

nofishonfriday:
Oh dear, seems SWS constantly needs to catch up on REAPER development. 🗡️ 
Thanks for the report.

nofishonfriday:
@musicbynumbers
> -SWS zoom tool (can no longer draw a box)

Can't confirm here. Ok, the zoom box is offset (due to the spacers, needs recalculation), but I can still draw the zoom boxes with some spacers inserted on initial testing.
Do you experience s.th. different? If so, can upload an example project?

![spacers](https://github.com/reaper-oss/sws/assets/7658194/df58a6ee-66b6-4e0c-a915-4e0e83bdca57)

musicbynumbers:
Hi @nofishonfriday -  I'll try and do a test project later but the zoom issue was actually trying to zoom into a lower set of items on say track 100 (big project) whilst also having groups of tracks above this that had track spacers roughly 10 tracks apart.

This seemed to cause the zoom of a few tracks/items at around the 100th track area to be off due to say about 10 spacers worth of spacers being spread out in the 100 or so tracks above them.

I removed the spacers from the project I noticed it in so will try to recreate.

musicbynumbers:
Try this reaper project. I've put some midi items right at the bottom. If I try to zoom into one of them (on a single track). For me, it will zoom vertically to 2 or 3 tracks and you should be able to see that the zoom area box whilst drawing the zoom area is off too. Let me know if it doesn't work and maybe it's worth me sending settings over etc :)

[SWS zoom test.zip](https://github.com/reaper-oss/sws/files/11563794/SWS.zoom.test.zip)

musicbynumbers:
Hi @nofishonfriday - just checking in that you saw the project I uploaded here and if you were able to reproduce it? 

No rush, it was just in case you didn't see it in the first place. :)

nofishonfriday:
[...]

PR #1787, opened 2023-10-26
Add support for TCP/MCP track spacers when acquiring context/track from mouse position

jeremybernstein:
GetTrackHeight return value includes spacer height.

Issue #1761

### DIFF
--- a/Breeder/BR_MouseUtil.cpp
+++ b/Breeder/BR_MouseUtil.cpp
@@ -56,7 +56,7 @@ const int MIDI_WND_UNKNOWN      = 3;
 /******************************************************************************
 * Helper functions                                                            *
 ******************************************************************************/
-static MediaTrack* GetTrackAreaFromY (int y, int* offset)
+static MediaTrack* GetTrackAreaFromY (int y, int* offset, int* spacerSize)
 {
 	/* Check if Y is in some TCP track or it's envelopes, *
 	*  returned offset is always for returned track       */
@@ -66,10 +66,15 @@ static MediaTrack* GetTrackAreaFromY (int y, int* offset)
 
 	int trackEnd = 0;
 	int trackOffset = 0;
+	int trackSpacerSize = 0;
 	for (int i = 0; i <= GetNumTracks(); ++i)
 	{
 		MediaTrack* currentTrack = CSurf_TrackFromID(i, false);
 		int height = *(int*)GetSetMediaTrackInfo(currentTrack, "I_WNDH", NULL);
+
+		trackSpacerSize = GetTrackSpacerSize(currentTrack);
+		height += trackSpacerSize;
+
 		if (currentTrack == master && TcpVis(master))
 			height += GetMasterTcpGap();
 		trackEnd += height;
@@ -80,17 +85,20 @@ static MediaTrack* GetTrackAreaFromY (int y, int* offset)
 			break;
 		}
 		trackOffset += height;
+		trackSpacerSize = 0;
 	}
 
 	WritePtr(offset, (track) ? (trackOffset) : (0));
+	WritePtr(spacerSize, trackSpacerSize);
 	return track;
 }
 
-static MediaTrack* GetTrackFromY (int y, int* trackHeight, int* offset)
+static MediaTrack* GetTrackFromY (int y, int* trackHeight, int* offset, int* spacerSize)
 {
 	int trackOffset = 0;
 	int trackH = 0;
-	MediaTrack* track = GetTrackAreaFromY(y, &trackOffset);
+	int spacer = 0;
+	MediaTrack* track = GetTrackAreaFromY(y, &trackOffset, &spacer);
 	if (track)
 	{
 		trackH = GetTrackHeight(track, NULL);
@@ -98,8 +106,9 @@ static MediaTrack* GetTrackFromY (int y, int* trackHeight, int* offset)
 			track = NULL;
 	}
 
-	WritePtr(trackHeight, (track) ? (trackH)      : (0));
-	WritePtr(offset,      (track) ? (trackOffset) : (0));
+	WritePtr(trackHeight,  track ? trackH             : 0);
+	WritePtr(offset,       track ? trackOffset        : 0);
+	WritePtr(spacerSize,   (track && spacer) ? spacer : 0);
 	return track;
 }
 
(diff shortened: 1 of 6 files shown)


Breeder/BR_Util.cpp
```diff
@@ -2838,12 +2864,16 @@ MediaTrack* HwndToTrack (HWND hwnd, int* hwndContext, POINT ptScreen)
 					if (!GetMediaTrackInfo_Value(chktrack, "B_SHOWINTCP"))
 						continue;
 
-					const double ypos = GetMediaTrackInfo_Value(chktrack, "I_TCPY"),
-					             h    = GetMediaTrackInfo_Value(chktrack, "I_TCPH");
+					const int spacerHeight = GetTrackSpacerHeight(chktrack);
```
cfillion, line 2867: Would be clearer named `spacerSize` / `GetTrackSpacerSize` since it's width in the MCP.

Breeder/BR_Util.cpp
```diff
@@ -2233,6 +2236,28 @@ int GetTrackEnvHeight (TrackEnvelope* envelope, int* offsetY, bool drawableRange
 	return envelopeHeight;
 }
 
+int GetTrackSpacerHeight (MediaTrack* track)
+{
+	// maybe cache the app version in a global var to avoid an API call here
+	// and in other places where we perform this check
+    bool trackSpacerIsSupported = atof(GetAppVersion()) >= 7 ? true : false;
```
cfillion, line 2243: That ternary is redundant as `>=` already gives the desired boolean so it's redundant.

Breeder/BR_MouseUtil.cpp
```diff
@@ -80,26 +85,30 @@ static MediaTrack* GetTrackAreaFromY (int y, int* offset)
...
 		trackH = GetTrackHeight(track, NULL);
 		if (y < trackOffset || y >= trackOffset + trackH)
 			track = NULL;
 	}
 
-	WritePtr(trackHeight, (track) ? (trackH)      : (0));
-	WritePtr(offset,      (track) ? (trackOffset) : (0));
+	WritePtr(trackHeight,  (track) ? (trackH)           : (0));
+	WritePtr(offset,       (track) ? (trackOffset)      : (0));
+	WritePtr(spacerHeight, (track && spacer) ? (spacer) : (0));
```
cfillion, line 111: Opportunity to cleanup the old code:

```suggestion
	WritePtr(trackHeight,  track ? trackH             : 0);
	WritePtr(offset,       track ? trackOffset        : 0);
	WritePtr(spacerHeight, (track && spacer) ? spacer : 0);
```
Last parenthesis is also unnecessary but I guess it's fine for better readability. 🤷

cfillion (commented):
Looking good! MCP and TCP seems to work perfectly.

Arrange view is incorrect if the first track has a spacer (and the master track is hidden). `TranslatePointToArrangeScrollY`'s 0 point gets offset by one spacer height by the scrollbar (`si.nPos`).

Breeder/BR_Util.cpp
```diff
@@ -2233,6 +2236,28 @@ int GetTrackEnvHeight (TrackEnvelope* envelope, int* offsetY, bool drawableRange
...
+    int returnValue = 0;
+
+    if (trackSpacerIsSupported)
+    {
+		// this is slightly inefficient to acquire the prefs value each time, but it could change
```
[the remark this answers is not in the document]
jeremybernstein: What about the case where the user changes the gap size mid-run? This was the (unpleasant) reason I perform the lookup each time, but maybe there's already a mechanism for knowing if the prefs change?

Breeder/BR_Util.cpp
jeremybernstein: roger that

Breeder/BR_Util.cpp
jeremybernstein: I think I got it I think

Breeder/BR_MouseUtil.cpp
jeremybernstein: 👍 I made these changes and reverted them because I didn't want to fix it everywhere, but I guess local gardens are nice, too.

Breeder/BR_Util.cpp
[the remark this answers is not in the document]
cfillion: `ConfigVar` holds a the pointer to REAPER's global variable not its value. The snippet above accesses the current value every time:
 
https://github.com/reaper-oss/sws/blob/c5723a718945bf42e8b924f165c698223cf9bb38/Utility/configvar.h#L50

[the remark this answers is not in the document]
jeremybernstein: OK I see, thank you!

Breeder/BR_Util.h
```diff
@@ -262,7 +263,7 @@ HWND GetTcpTrackWnd (MediaTrack* track, bool &isContainer);
 HWND GetNotesView (HWND midiEditor);
 HWND GetPianoView (HWND midiEditor);
 HWND GetTrackView (HWND midiEditor);
-MediaTrack* HwndToTrack (HWND hwnd, int* hwndContext, POINT ptScreen);  // context: 0->unknown, 1->TCP, 2->MCP (works even if hwnd is not a track but something else in mcp/tcp). 
+MediaTrack* HwndToTrack (HWND hwnd, int* hwndContext, POINT ptScreen);  // (context & 0x0F): 0->unknown, 1->TCP, 2->MCP (works even if hwnd is not a track but something else in mcp/tcp). If the point is over a track spacer, (context & 0x10) == 0x10
```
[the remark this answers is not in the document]
jeremybernstein: Roger that

cfillion (approved):
Looks good to me and seems to work well. Could you squash the commits?

jeremybernstein:
@cfillion done.

jeremybernstein:
> Testing the SWS beta bulid.
> 
> This action is not working properly: **SWS/wol: Adjust envelope or track height under mouse cursor, zoom center to mouse cursor (MIDI CC relative/mousewheel)**
> 
> If i want to resize a track that is **just below** a track spacer, it only increases the track height with mousewheel up, but with mousewheel down, it also tries to increase the height in minimal steps, and after it gets resized i can't even get back to normal size, even with built in commands like: **View: Toggle track zoom to minimum height**
> 
> It works correctly if i assign a Midi CC, it only has this error with mousewheel down.

I am seeing issues with this action with or without track spacers. We can look into it, thanks for the report.

PR #1792, opened 2023-11-06
Use the more adequate GoToRegion API method in Region Playlist + other improvements 

### LINKED ISSUE #1512: region playlist nested markers and regions
Not sure if this is a bug or feature ....

I have various regions defined with markers within them.  

Regions and markers

When I add the regions to the playlist I get this  warning

Playlist warning

I use markers a lot in my project and would rather not have to delete them in order to use the region playlist.  Is there a way round this?

The regions also don't play properly.  Often playing until the reach the first marker then moving on to the next region

Szapi:
Fixes  #1512

My goal was to make the Region Playlist more robust and reliable. In my opinion, in its current state it's just not good enough for any serious tasks. I really wanted to use it with my band for live gigs and rehearsals, but the issue with markers and the regular sync losses were unacceptable.

Through experiments I found that proper bound checking was needed to avoid the occasional sync loss/skipped regions due to numeric error. Even though reaper says a region starts at t1 and ends at t2 (EnumProjectMarkers), the reported play position (from GetPlayPosition) can be slightly less than t1 or slightly greater than t2. I chose an epsilon of 10ms as this value was already used in a couple of places.

A much more common bug was the sync loss after the very first region, where it just kept playing past the end of the region. This was due to a timing issue. I could reproduce it with both the smooth seek method and the GoToRegion method. My measurements clearly indicated a correlation between the occurrence of this bug and the time between the PlaylistPlay function call and the first PlaylistRun callback. If this measured time was less than about 5ms, the bug almost always happened. It seems like if not enough time has passed between the OnPlayButton call and the 'scheduling' of the second region via smooth seek/GoToRegion, then Reaper ignored the schedule 'request'. After I delayed the scheduling by 5 callbacks (160-200 ms), I couldn't reproduce this bug anymore. I added this small delay to all GoToRegion calls, just to be safe (and also it was easier this way).

Sadly the ending of the playlist must be done with a smooth seek, which of course also considers regular markers. I timed this after any markers that are within a region, but to be safe, there should be at least 0.5 seconds between the end of the region and any markers within.

I added two checks with warning dialogs due to these changes.

Also I stopped the SYNC LOSS text from flashing when the playlist finished, to make it look more professional.

I tried to keep the sync loss and end-of-playlist behavior the same as before. In my manual testing it was OK.

I tested my changes with looped/missing/touching regions as well as some loop corner cases I found in the c++ code comments. I ran a quite complex playlist with all these different scenarios on infinite loop, with a very elegant abort() call on sync loss or skipped regions, and it ran for about 250 loops without failure before I shut it down.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,6 @@
 
 # External files
 /vendor
+
+# IDE's
+/.vscode
--- a/.gitignore
+++ b/.gitignore
@@ -5,5 +5,3 @@
 # External files
 /vendor
 
-# IDE's
-/.vscode
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,6 +89,8 @@ if(MSVC)
     # https://web.archive.org/web/20151123133638/https://connect.microsoft.com/VisualStudio/feedback/details/1789709/visual-c-2015-runtime-broken-on-windows-server-2003-c-11-magic-statics
     /Zc:threadSafeInit-
   )
+elseif(APPLE)
+  add_compile_options(-fsigned-char -fstack-protector-strong -fdiagnostics-color --sysroot ${CMAKE_OSX_SYSROOT})
 else()
   add_compile_options(-fsigned-char -fstack-protector-strong -fdiagnostics-color)
 endif()
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -89,8 +89,6 @@ if(MSVC)
     # https://web.archive.org/web/20151123133638/https://connect.microsoft.com/VisualStudio/feedback/details/1789709/visual-c-2015-runtime-broken-on-windows-server-2003-c-11-magic-statics
     /Zc:threadSafeInit-
   )
-elseif(APPLE)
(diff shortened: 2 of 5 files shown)


Buy-One:
Have these changes been added to the official build?

Szapi:
Nah. I just proposed these changes, but the original authors still need to review and accept it.
I don't know when this will happen.

Buy-One:
OK

.gitignore
```diff
@@ -4,3 +4,6 @@
 
 # External files
 /vendor
+
+# IDE's
+/.vscode
```
cfillion, line 9: Editor preferences belongs in the user's own git configuration, not the repository's.

git has a per-repository user config file at `.git/info/exclude` and user-wide at `~/.config/git/ignore`.

cfillion (commented):
Don't have time to do a in-depth review and testing right now, but it would be better to have the commits squashed where each one is a standalone, complete step (that could be released by itself). For example: a commit refactoring a thing, another building upon that to add a feature, another fixing a loosely-related bug.

Szapi:
Oh. Okay. Sorry, Git and Github is still new to me, I only have experience with centralized VCS's. Thanks for the infos.

Szapi:
I'll close this PR and create a new one with squashed commits.

PR #1795, opened 2023-11-11
New Lua timer functions for development purposes

Szapi:


cfillion:
Looking for something like this? https://forum.cockos.com/showthread.php?t=283461 😉

Szapi:
Ooh. Okay, I found the ActionTimer, but not this 😄 in this case you can just reject this PR. It was a nice little introduction to the Lua interface though.

Ah. I didn't realize the C++11 STL is forbidden. 😕 A real bummer.

PR #1801, opened 2023-11-18
Speed up SWS Track Notes

### LINKED ISSUE #1614: Request : make SWS notes fast
I notice that the SWS notes functions can get quite slow. Even simply writting things in the SWS notes is lagging in my project at this point. 

I use them a lot, they are very cool. But if I use "Script: cfillion_Duplicate tracks (preserving SWS track notes).lua" with folders (lots of tracks, 50+), the limit of current SWS notes start to show. The action takes half a minute to do its work. Heavy lifting it seems... But maybe it could be made lighter ? 

So I guess my request would be to think about making SWS notes a bit faster in the long run, for people like me that have started to use them as a very important tool. Not only to mark a few things ("change this, eq that, sample this"), but rather as an elaborate track tagging system, notes that can influence the actions of external scripts. 

SWS notes are VERY powerful. Make them go... faster please :p.
xxxxxxxx
louis

Szapi:
Fixes #1614

Calling GuidToTrack in for-loops made some operations painfully slow, e.g. "cfillion_Duplicate tracks (preserving SWS track notes).lua".

I made track-to-tracknotes lookups much faster by just comparing project pointers and guids, which is sufficient to globally identify any track.

Duplicating 256 tracks with notes:

Before:
GetSWSTrackNotes took 888 ms
Duplicate tracks took 41 ms
SetSWSTrackNotes took 4692 ms
The entire lua script took 5637 ms

Now:
GetSWSTrackNotes took: 1 ms
Duplicate tracks took: 53 ms
SetSWSTrackNotes took: 2 ms
The entire lua script took: 67 ms

It's barely even measureable in ms which made me worried that I screwed up something, but it seems to be working by my testing.

### DIFF
--- a/SnM/SnM_Find.cpp
+++ b/SnM/SnM_Find.cpp
@@ -115,9 +115,10 @@ bool TrackNotesMatch(MediaTrack* _tr, const char* _searchStr)
 	bool match = false;
 	if (_tr)
 	{
+		const MediaTrackID trId = TrackToTrackID(_tr);
 		for (int i=0; i < g_SNM_TrackNotes.Get()->GetSize(); i++)
 		{
-			if (g_SNM_TrackNotes.Get()->Get(i)->GetTrack() == _tr)
+			if (g_SNM_TrackNotes.Get()->Get(i)->GetTrackID() == trId)
 			{
 				match = stristr(g_SNM_TrackNotes.Get()->Get(i)->GetNotes(), _searchStr) != NULL;
 				break;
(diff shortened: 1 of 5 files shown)


cfillion (commented):
I like this `MediaTrackID` idea of a track handle object! In this case it's a bit overkill because all comparisons are done with tracks of the current project against track notes of the current project, so a simple GuidEquals in a loop is enough. The identical loops can be extracted for code reuse.

I made an simplified version of this PR here: #1835.

PR #1802, opened 2023-11-19
Use the more adequate GoToRegion API method in Region Playlist + stability improvements

### LINKED ISSUE #914: Region Playlist - No stopping allowed?
Hello good people!

I'm trying to use a Region Playlist to sequence songs in a live show.     Looks like it is working great but there's no way to stopping/pausing the playlist in between regions.   I tried adding a marker with an "!1016" (and other action numbers) at the end of my regions, nope...  The playlist just won't stop...

Now, would there be a way of doing this?   Maybe there is, I just don't see how.    Could it be a "feature request"?    Say, if we put "0" on the Loop count column, the Playlist could stop after the current region has been played.   That'd be simple and awesome for designing set lists in a live show.

Thanks for all the hard work, you guys rock!

Marc aka Khorus

### LINKED ISSUE #1512: region playlist nested markers and regions

### LINKED ISSUE #1575: FR: Support for action markers in Region Playlist OR an action to stop playback
If playback was launched with Region Playlist transport it can't be stopped/paused with action markers.

So my suggestion is to either add support for action markers or alternatively (which is presumably a less labor intensive solution), to add action **SWS/S&M: Region Playlist - Stop**, because there's already one for starting playback **SWS/S&M: Region Playlist - Play** and accompanying it with a counterpart for Stop seems logical.

Szapi:
Fixes  #1512
Fixes #1575
Fixes  #914

My goal was to make the Region Playlist more robust and reliable. In my opinion, in its current state it's just not good enough for any serious tasks. I really wanted to use it with my band for live gigs and rehearsals, but the issue with markers and the regular sync losses were unacceptable.

Through experiments I found that proper bound checking was needed to avoid the occasional sync loss/skipped regions due to numeric error. Even though reaper says a region starts at t1 and ends at t2 (EnumProjectMarkers), the reported play position (from GetPlayPosition) can be slightly less than t1 or slightly greater than t2. I chose an epsilon of 10ms as this value was already used in a couple of places.

A much more common bug was the sync loss after the very first region, where it just kept playing past the end of the region. This was due to a timing issue. I could reproduce it with both the smooth seek method and the GoToRegion method. My measurements clearly indicated a correlation between the occurrence of this bug and the time between the PlaylistPlay function call and the first PlaylistRun callback. If this measured time was less than about 5ms, the bug almost always happened. It seems like if not enough time has passed between the OnPlayButton call and the 'scheduling' of the second region via smooth seek/GoToRegion, then Reaper ignored the schedule 'request'. After I delayed the scheduling by 5 callbacks (160-200 ms), I couldn't reproduce this bug anymore. I added this small delay to all GoToRegion calls, just to be safe (and also it was easier this way).

Sadly the ending of the playlist must be done with a smooth seek, which of course also considers regular markers. I timed this after any markers that are within a region, but to be safe, there should be at least 0.5 seconds between the end of the region and any markers within.

I added two checks with warning dialogs due to these changes.

Also I stopped the SYNC LOSS text from flashing when the playlist finished, to make it look more professional.

I tried to keep the sync loss and end-of-playlist behavior the same as before. In my manual testing it was OK.

I tested my changes with looped/missing/touching regions as well as some loop corner cases I found in the c++ code comments. I ran a quite complex playlist with all these different scenarios on infinite loop, with a very elegant abort() call on sync loss or skipped regions, and it ran for about 250 loops without failure before I shut it down.

### DIFF
--- a/SnM/SnM_RegionPlaylist.cpp
+++ b/SnM/SnM_RegionPlaylist.cpp
@@ -1644,6 +1644,11 @@ void PlaylistRun()
 		}
 	}
 
+	if (-1 == g_playCur && g_unsync) {
+		// Stop the 'SYNC LOSS' text from flashing after playlist ended. Looks unprofessional.
+		updated = false;
+	}
+
 	g_lastRunPos = pos;
 	if (updated && (g_osc || g_rgnplWndMgr.Get()))
 	{
--- a/SnM/SnM_RegionPlaylist.cpp
+++ b/SnM/SnM_RegionPlaylist.cpp
@@ -1397,8 +1397,7 @@ void PrepareToEndPlaylist ()
 	EnumProjectMarkers2(NULL, markerIdx, nullptr, &g_safeTimeToEndPlaylist, nullptr, nullptr, nullptr);
 }
 
-// Should only be called from SeekRegionDelayer and SeekRegion
-void SeekRegionImpl (const int _reaperRegionId)
+void SeekRegion (const int _reaperRegionId)
 {
 	const double cursorpos = GetCursorPositionEx(NULL);
 	PreventUIRefresh(1);
@@ -1413,49 +1412,6 @@ void SeekRegionImpl (const int _reaperRegionId)
 	snprintf(dbg, sizeof(dbg), "GoToRegion() - Reaper Region Id: %d\n", _reaperRegionId);
 	OutputDebugString(dbg);
 #endif
-}
-
-// GoToRegion timing issue workaround
-class SeekRegionDelayer {
-	static constexpr int delayAmount = 5;
-
-	int regionId;
-	int currentDelay = -1;
-
-public:
-	void Schedule (int _regionToSchedule)
-	{
-		regionId = _regionToSchedule;
-		currentDelay = delayAmount;
-	}
-
-	void Reset ()
-	{
-		currentDelay = -1;
-	}
-
-	void Update ()
-	{
-		if (currentDelay > 0)
-			--currentDelay;
-		
-		if (currentDelay == 0) {
-			currentDelay = -1;
-			SeekRegionImpl(regionId);
-		}
-	}
-};
-
-SeekRegionDelayer g_seekRegionDelayer;
-
-void SeekRegion (const int _reaperRegionId)
-{
-	if (g_playPlaylist<0) {
-		// Called from PlaylistPlay to play the very first region --> must not delay
-		SeekRegionImpl(_reaperRegionId);
-	} else {
-		g_seekRegionDelayer.Schedule(_reaperRegionId);
-	}
 }
 
 enum class SeekMethod {
@@ -1509,7 +1465,8 @@ constexpr double timeEps = 0.01; // 10 ms
 
 static bool IsInCurrentRegion (const double pos)
 {
-	return (g_curRgnPos-timeEps) < pos && pos < (g_curRgnEnd+timeEps);
+	// not subtracting timeEps from pos to avoid occasionally skipped seeks (#886)
+	return g_curRgnPos < pos && pos < (g_curRgnEnd+timeEps);
 }
 
 static bool IsInNextRegion (const double pos)
@@ -1518,11 +1475,11 @@ static bool IsInNextRegion (const double pos)
 		return IsInCurrentRegion(pos);
 
 	// We need to be careful because the relaxed intervals
-	// (g_curRgnPos  - timeEps, g_curRgnEnd  + timeEps)
-	// (g_nextRgnPos - timeEps, g_nextRgnEnd + timeEps)
+	// (g_curRgnPos,  g_curRgnEnd  + timeEps)
+	// (g_nextRgnPos, g_nextRgnEnd + timeEps)
 	// can overlap if the regions are very close or touching
 	return !IsInCurrentRegion(pos) &&
-			(g_nextRgnPos-timeEps) < pos && pos < (g_nextRgnEnd+timeEps);
+		g_nextRgnPos < pos && pos < (g_nextRgnEnd+timeEps);
 }
 
 // the meat!
@@ -1533,8 +1490,6 @@ void PlaylistRun()
 {
 	if (g_playPlaylist<0)
 		return;
-
-	g_seekRegionDelayer.Update();
 
 #if defined(_SNM_RGNPL_DEBUG1) || defined(_SNM_RGNPL_DEBUG2)
 	char dbg[256] = "";
@@ -1727,7 +1682,7 @@ void PlaylistPlay(int _plId, int _itemId)
 		}
 
 		const int unsafeRegion = pl->GetRegionWithUnsafeMarker();
-		if (unsafeRegion>0)
+		if (unsafeRegion >= 0)
 		{
 			char msg[512] = "";
 			snprintf(msg, sizeof(msg),
@@ -1740,7 +1695,7 @@ void PlaylistPlay(int _plId, int _itemId)
 		}
 
 		const int dangerouslyShortRegion = pl->GetDangerouslyShortRegion();
-		if (dangerouslyShortRegion>0)
+		if (dangerouslyShortRegion >= 0)
 		{
 			char msg[256] = "";
 			snprintf(msg, sizeof(msg),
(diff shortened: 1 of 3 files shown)


SnM/SnM_RegionPlaylist.cpp
```diff
@@ -485,13 +536,15 @@ void GetMonitoringInfo(WDL_FastString* _curNum, WDL_FastString* _cur,
...
 			else
 			{
-				int id, idx = FindMarkerRegion(NULL, GetPlayPositionEx(NULL), SNM_REGION_MASK, &id);
-				if (id > 0)
-				{
-					char buf[64]="";
-					EnumMarkerRegionDesc(NULL, idx, buf, sizeof(buf), SNM_REGION_MASK, false, true, false);
-					_curNum->SetFormatted(16, "%d", GetMarkerRegionNumFromId(id));
-					_cur->Set(buf);
+				if (-1 != g_playCur) {
```
Szapi, line 539: Oops. This was not equivalent, but at least now when the playlist ends, the current region display immediately goes blank. Oh well.

SnM/SnM_RegionPlaylist.cpp
```diff
@@ -1326,23 +1379,102 @@ int GetPrevValidItem(int _plId, int _itemId, bool _startWith, bool _repeat, bool
...
+#if defined(_SNM_RGNPL_DEBUG1) || defined(_SNM_RGNPL_DEBUG2)
+	char dbg[128] = "";
+	snprintf(dbg, sizeof(dbg), "GoToRegion() - Reaper Region Id: %d\n", _reaperRegionId);
+	OutputDebugString(dbg);
+#endif
+}
+
+// GoToRegion timing issue workaround
+class SeekRegionDelayer {
+	static constexpr int delayAmount = 5;
```
Szapi, line 1420: Maybe even a single delay would be enough, but I wouldn't go below 3, just to be safe.

cfillion (commented):
What is the reasoning behind the delayer? Under what conditions is it required? (Comment only says "GoToRegion timing issue workaround".)

I've tried stress-testing this, it makes playlists more vulnerable to polling rate slowdowns. For instance, the jump to the next region after a 0.5s-long one occurs a few seconds late if the defer cycle rate is brought down artificially using this script:

```lua
local function loop()
  for i = 1, 1<<20 do reaper.GetTrack(nil, 0) end
  reaper.defer(loop)
end

reaper.defer(loop)
```

Whereas the previous code (or new w/ delayer bypassed) correctly jumps to the next region virtually immediately. (In the real world such slowdowns can happen if the user runs multiple unoptimized scripts, visible plugin UIs, older PC...)

Before:
Before

After this PR:
After this PR

Szapi:
I tried to explain the delayer in the PR description. I have a theory, though I might be completely wrong 🙂

Without the delay, the second region was 'scheduled' on the very first PlaylistRun callback after PlaylistPlay.

The PlaylistRun method is continuously being called by reaper with ~30Hz. If you lay out these callbacks on a timeline, that creates a grid with ~33,3ms gaps between the grid lines. When you start the playlist, the PlaylistRun method call will randomly (maybe not so randomly, but that depends on Reaper backend, so we don't know that) fall into one of these 33ms windows. If it falls too close to the end of the window, meaning the next PlaylistRun will be called almost instantly, then there's a very good chance, that we will go out of sync, and just continue playing past the end of the first region, without jumping to the second region. This happened both with the smooth seek and the GoToRegion implementation as well. I guess they both use the same logic under the hood in Reaper, with the exception that GoToRegion will ignore regular markers.

This bug occured very rarely, maybe like 1% of the cases. I'm not sure. But when I measured the time between PlaylistPlay and PlaylistRun calls, I could clearly see a correlation between the bug happening and the measured time. If it was ~5ms or less, the bug almost always happened.

After I delayed the GoToRegion/smooth seek, I could not reproduce this bug anymore.

I hope this clears it up. We can lower the delay to a single cycle if that solves the stress test issues.

cfillion:
Thanks! I think I've narrowed it down to the 10ms subtraction from region start causing random calls to `GoToRegion` too soon without the delayer.

Can you give the patch below a try? On my system it solves the seek issues without the delayer when the timer runs at the nominal 30ms rate and when slowed down in a heavy session (or simulated). (For completeness I've also tried with it sped up to [16ms](https://cfillion.ca/files/reaper/60fps/))

If there are still bugs with that, could you share a RPP with the regions & playlist data? Here's the one I used: [Untitled.RPP.txt](https://github.com/reaper-oss/sws/files/14150745/Untitled.RPP.txt).

<details><summary>Patch</summary>

```diff
diff --git a/SnM/SnM_RegionPlaylist.cpp b/SnM/SnM_RegionPlaylist.cpp
index 0ee6f5a8..14b3e7b4 100644
--- a/SnM/SnM_RegionPlaylist.cpp
+++ b/SnM/SnM_RegionPlaylist.cpp
@@ -1397,8 +1397,7 @@ void PrepareToEndPlaylist ()
 	EnumProjectMarkers2(NULL, markerIdx, nullptr, &g_safeTimeToEndPlaylist, nullptr, nullptr, nullptr);
 }
 
-// Should only be called from SeekRegionDelayer and SeekRegion
-void SeekRegionImpl (const int _reaperRegionId)
+void SeekRegion (const int _reaperRegionId)
 {
 	const double cursorpos = GetCursorPositionEx(NULL);
 	PreventUIRefresh(1);
@@ -1415,49 +1414,6 @@ void SeekRegionImpl (const int _reaperRegionId)
 #endif
 }
 
-// GoToRegion timing issue workaround
-class SeekRegionDelayer {
-	static constexpr int delayAmount = 5;
-
-	int regionId;
-	int currentDelay = -1;
-
-public:
-	void Schedule (int _regionToSchedule)
-	{
-		regionId = _regionToSchedule;
-		currentDelay = delayAmount;
-	}
-
-	void Reset ()
-	{
-		currentDelay = -1;
-	}
-
-	void Update ()
-	{
-		if (currentDelay > 0)
-			--currentDelay;
-		
-		if (currentDelay == 0) {
-			currentDelay = -1;
-			SeekRegionImpl(regionId);
-		}
-	}
-};
-
-SeekRegionDelayer g_seekRegionDelayer;
-
-void SeekRegion (const int _reaperRegionId)
-{
-	if (g_playPlaylist<0) {
-		// Called from PlaylistPlay to play the very first region --> must not delay
-		SeekRegionImpl(_reaperRegionId);
-	} else {
-		g_seekRegionDelayer.Schedule(_reaperRegionId);
-	}
-}
-
 enum class SeekMethod {
 	IgnoreMarkers,
 	ConsiderMarkers
@@ -1509,7 +1465,7 @@ constexpr double timeEps = 0.01; // 10 ms
 
 static bool IsInCurrentRegion (const double pos)
 {
-	return (g_curRgnPos-timeEps) < pos && pos < (g_curRgnEnd+timeEps);
+	return g_curRgnPos < pos && pos < (g_curRgnEnd+timeEps);
 }
 
 static bool IsInNextRegion (const double pos)
@@ -1522,7 +1478,7 @@ static bool IsInNextRegion (const double pos)
 	// (g_nextRgnPos - timeEps, g_nextRgnEnd + timeEps)
 	// can overlap if the regions are very close or touching
 	return !IsInCurrentRegion(pos) &&
-			(g_nextRgnPos-timeEps) < pos && pos < (g_nextRgnEnd+timeEps);
+			g_nextRgnPos < pos && pos < (g_nextRgnEnd+timeEps);
 }
 
 // the meat!
@@ -1534,8 +1490,6 @@ void PlaylistRun()
 	if (g_playPlaylist<0)
 		return;
 
-	g_seekRegionDelayer.Update();
-
 #if defined(_SNM_RGNPL_DEBUG1) || defined(_SNM_RGNPL_DEBUG2)
 	char dbg[256] = "";
 #endif
@@ -1777,7 +1731,6 @@ void PlaylistPlay(int _plId, int _itemId)
 			g_endOfPlaylistSeekIssued = false;
 			if (SeekItem(_plId, _itemId, (g_playPlaylist==_plId ? g_playCur : -1), SeekMethod::IgnoreMarkers))
 			{
-				g_seekRegionDelayer.Reset();
 				g_playPlaylist = _plId; // enables PlaylistRun()
 				if (RegionPlaylistWnd* w = g_rgnplWndMgr.Get())
 					w->Update(); // for the play button, next/previous region actions, etc....
```
</details>

PR #1842, opened 2024-02-11
Add GetTrackHeightAndOffsetWithSpacers()

jeremybernstein:
GetTrackHeight() is now balanced to SetTrackHeight(), as previously -- it will get the 'content' portion height and offset of the track in the TCP. If you need the true track offset (- spacerSize) or height (+ spacerSize), use GetTrackHeightAndOffsetWithSpacers().

Updated a couple of Zoom.cpp functions to take the spacer size into account, missed on the first pass.

### DIFF
--- a/SnM/SnM_Window.cpp
+++ b/SnM/SnM_Window.cpp
@@ -233,8 +233,7 @@ void GetVisibleTCPTracks(WDL_PtrList<MediaTrack>* _trList)
 	{
 		MediaTrack* tr = CSurf_TrackFromID(i, false);
 		TCPY = static_cast<int>(GetMediaTrackInfo_Value(tr, "I_TCPY"));
-		TCPH = static_cast<int>(GetMediaTrackInfo_Value(tr, "I_TCPH"));
-		GetTrackGap(TCPH, &topGap, &bottomGap);
+		TCPH = GetTrackHeight(tr, nullptr, &topGap, &bottomGap);
 		if ((TCPY + TCPH - bottomGap > 0) && (TCPY + topGap < arrangeHeight))
 			_trList->Add(tr);
 	}
--- a/Zoom.cpp
+++ b/Zoom.cpp
@@ -170,9 +170,12 @@ void VertZoomRange(int iFirst, int iNum, bool* bZoomed, bool bMinimizeOthers, bo
 				{
 					int trackHeight = 0;
 					if (obeyHeightLock && locked)
+					{
 						trackHeight = static_cast<int>(GetMediaTrackInfo_Value(tr, "I_HEIGHTOVERRIDE"));
+						trackHeight += GetTrackSpacerSize(tr);
+					}
 					else
-						trackHeight = GetTrackHeightFromVZoomIndex(tr, 0);
+						trackHeight = GetTrackHeightWithSpacer(tr);
 
 					trackHeight += CountTrackEnvelopePanels(tr) * GetEnvHeightFromTrackHeight(trackHeight);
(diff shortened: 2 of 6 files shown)


SnM/SnM_Window.cpp
```diff
@@ -232,9 +232,7 @@ void GetVisibleTCPTracks(WDL_PtrList<MediaTrack>* _trList)
 	for (int i = 1; i <= GetNumTracks(); i++) // skip master
 	{
 		MediaTrack* tr = CSurf_TrackFromID(i, false);
-		TCPY = static_cast<int>(GetMediaTrackInfo_Value(tr, "I_TCPY"));
-		TCPH = static_cast<int>(GetMediaTrackInfo_Value(tr, "I_TCPH"));
-		GetTrackGap(TCPH, &topGap, &bottomGap);
+		TCPH = GetTrackHeightAndOffsetWithSpacers(tr, &TCPY, &topGap, &bottomGap);
```
cfillion, line 235: This breaks the "Toggle offscreen item selection" actions' toggle state by changing `TCPY` from relative (to the visible area) to absolute (`I_TCPY + si.nPos`). Also spacers should not be included here.

```diff
diff --git a/SnM/SnM_Window.cpp b/SnM/SnM_Window.cpp
index eae2460c..c41e5bbb 100644
--- a/SnM/SnM_Window.cpp
+++ b/SnM/SnM_Window.cpp
@@ -232,7 +232,8 @@ void GetVisibleTCPTracks(WDL_PtrList<MediaTrack>* _trList)
 	for (int i = 1; i <= GetNumTracks(); i++) // skip master
 	{
 		MediaTrack* tr = CSurf_TrackFromID(i, false);
-		TCPH = GetTrackHeightAndOffsetWithSpacers(tr, &TCPY, &topGap, &bottomGap);
+		TCPY = static_cast<int>(GetMediaTrackInfo_Value(tr, "I_TCPY"));
+		TCPH = GetTrackHeight(tr, nullptr, &topGap, &bottomGap);
 		if ((TCPY + TCPH - bottomGap > 0) && (TCPY + topGap < arrangeHeight))
 			_trList->Add(tr);
 	}
```

jeremybernstein: thanks, fixed and squash/force-pushed to branch

Zoom.cpp
```diff
@@ -223,10 +226,13 @@ void VertZoomRange(int iFirst, int iNum, bool* bZoomed, bool bMinimizeOthers, bo
 			if (bZoomed[i])
 			{
 				if (i + 1 == iNum)
-					iEachHeight +=leftOverHeight;
+					iEachHeight += leftOverHeight;
 				MediaTrack* tr = CSurf_TrackFromID(i + iFirst, false);
 				if (!obeyHeightLock || !GetMediaTrackInfo_Value(tr, "B_HEIGHTLOCK"))
+				{
 					GetSetMediaTrackInfo(tr, "I_HEIGHTOVERRIDE", &iEachHeight);
+					iEachHeight += GetTrackSpacerSize(tr);
```
cfillion, line 234: Each selected track should end up with the same height, this line increases it once per track: https://i.imgur.com/tdFgLGS.png

Pushing a fix. Rebased off master to test with the workaround for "minimize others" being completely broken in v6.76+.

Also renamed `GetTrackHeightAndOffsetWithSpacers` to `GetTrackHeightWithSpacer` because `GetTrackHeight` anonymously also does the `AndOffset` part (and this was getting long to type).

jeremybernstein: Thanks, sorry about that!

PR #1920, opened 2025-01-21
DoSelectFiles (Xenakios/SWS: Choose files for random insert...): fix multi-file return value processing

jeremybernstein:
Single-value processing was working. This is the only occurrence of BrowseForFiles() with multiple return values, as far as I could find.


cfillion:
Done in #1934.

PR #1980, opened 2025-10-15
dB calculation improvements

X-Raym:
- replace exp by pow for more accurate val from dB calculation

I saw that in some in my EEL scripts, exp based calculation that exp(NewVol*0.115129254); was not precise enough, compare to 
NewVolGain=pow(10, NewVol / 20.0);

Indeed, adding 1dB and removing 1dB with the exp(NewVol*0.115129254); method doesn't return the sound to original dB value (there is a tiny margin). That's a small detail, but as a perfect way to do the calculation can be done, it worths updating it.

I indeed took this method for my scripts 10 years ago and had to update about 25 scripts manually because I realize there was this error. A small insignificant imprecision in SWS, a code base of reference for ReaScripting, can lead to errors down the line.

Best,

### DIFF
--- a/Xenakios/ExoticCommands.cpp
+++ b/Xenakios/ExoticCommands.cpp
@@ -311,7 +311,8 @@ void DoNudgeItemVols(bool UseConf,bool Positive,double TheNudgeAmount)
 					NewVol=OldVolDB+NudgeAmount; else NewVol=OldVolDB-NudgeAmount;
 				double NewVolGain;
 				if (NewVol>-144.0)
-					NewVolGain=exp(NewVol*0.115129254);
+					NewVolGain=pow(10, NewVol / 20);
+
 				else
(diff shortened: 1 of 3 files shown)


cfillion:
Could you split the whitespace changes to its own commit?

cfillion:
1. `git reset HEAD~1` to undo the commit
2. `git add -p` and select only the whitespace change or the math change
3. `git commit`
4. Goto 2
5. `git push -f`

...but I'm now noticing that the PR was made from the master branch, meaning that yours will eventually diverge from upstream. Best to create the new commits in a new branch & send a new PR for that new branch (between steps 1&2: `git switch -c exp2pow`).

X-Raym:
see https://github.com/reaper-oss/sws/pull/1981

PR #2001, opened 2026-01-28
Add accessibility keyboard navigation toggle (Tab/arrows/typeahead)

ScottChesworth:
In the Groove tool dialog (and most others), SWS passes keystrokes through to REAPER's main window. That makes sense for most folks who split their input between mouse and keyboard, but for screen reader users, it makes navigation in those dialogs difficult. This PR adds "Enabled deeper keyboard navigation in SWS dialogs (improves accessibility)" in the SWS Options menu. For now I've used the Groove tool dialog as proof of concept, hoping I can extend this wider if the approach I've taken is suitable. 

Here are the changes to test in the Groove tool dialog once the new option has been enabled:

- Tab and Shift+Tab work to navigate through controls
- Arrows work to change radiobuttons
- - Typeahead works in the list box, making it quicker for users to jump to grooves they know the names of
- - Space clicks buttons when those have focus, plays/stops when focus is on any other type of control
- - Control+Z passes through when focus is in the list of grooves so users can apply, undo and retry

FYI I had help from Codex on this. Have been proofing its output best I can and user testing every step along the way. I'll be happy to round up other screen reader users for wider testing too. In particular I'm aware this is all Win32 specific right now and I don't know how SWELL will handle things on Mac. It's easy for me to reach VoiceOver users who can find out though.

For now, I'd like to get feedback before spending any more time on it. All thoughts welcome.

### DIFF
--- a/sws_wnd.cpp
+++ b/sws_wnd.cpp
@@ -560,6 +560,62 @@ int SWS_DockWnd::keyHandler(MSG* msg, accelerator_register_t* ctx)
 	SWS_DockWnd* p = (SWS_DockWnd*)ctx->user;
 	if (p && p->IsActive(true))
 	{
+		const HWND focused = GetFocus();
+		const bool focusInDockWnd = focused && (focused == p->m_hwnd || IsChild(p->m_hwnd, focused));
+		const auto allowSpaceToDialog = [focused]() -> bool {
+			if (!focused)
+				return false;
+			char className[64] = "";
+			GetClassName(focused, className, sizeof(className));
+			if (!strcmp(className, "ListBox"))
+				return false;
+			if (!strcmp(className, "Button"))
+			{
+				const LONG_PTR style = GetWindowLongPtr(focused, GWL_STYLE);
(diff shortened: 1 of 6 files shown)


sws_wnd.cpp
```diff
@@ -560,6 +560,62 @@ int SWS_DockWnd::keyHandler(MSG* msg, accelerator_register_t* ctx)
...
+						return true;
+				}
+			}
+			return true;
+		};
+
+		if (g_SNM_AccessibilityKeyboardNav && msg->message == WM_KEYDOWN)
+		{
+			const bool ctrlDown = (GetKeyState(VK_CONTROL) & 0x8000) != 0;
+			const bool altDown = (GetKeyState(VK_MENU) & 0x8000) != 0;
```
cfillion: SWELL only has `GetAsyncKeyState`. `BS_TYPEMASK`, `BS_CHECKBOX`, `BS_RADIOBUTTON` and `IsDialogMessage` are only available on Windows.

